PR #1, opened 2023-02-08
Seed repo with prototype.

joshualitt:
This CL includes the protoype bindings generator and the initial rough bindings. These bindings work for some trivial use cases, but there is still much work to do.


pubspec.lock
```diff
@@ -0,0 +1,365 @@
+# Generated by pub
```
kevmoo, line 1: we almost certainly want to exclude this!

kevmoo (changes requested):

pubspec.lock
joshualitt: Done.

kevmoo (approved):
LGTM

tool/bindings_generator/generate_bindings.mjs
```diff
@@ -0,0 +1,73 @@
...
+// TODO(joshualitt): Import js types properly.
+import 'dart:typed_data';
+import 'package:js/js.dart';
+`);
+}
+
+async function translateIDLs() {
+  const files = await idl.listAll();
+  const translator = new Translator();
+  for (const [shortname, file] of Object.entries(files)) {
```
sigmundch, line 32: Main suggestion (OK to do separately): we'd like to generate multiple libraries instead of one monolithic library. It helps with eng tooling (e.g. review tools cannot open files this big), but also with API discovery for users.

kevmoo: one mini-lib for each file the exports would be GREAT!

joshualitt: Let's do this in a follow on. Create #3.

joshualitt:
Okay, I'm going to land this so I can start work on #2 and #3. Please feel free to continue commenting on this CL and I will either address any issues in the follow on or file bugs.

tool/bindings_generator/generate_bindings.mjs
```diff
@@ -0,0 +1,73 @@
...
+
+function generateHeader() {
+  return (
+`// Copyright (c) ${util.year}, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// TODO(joshualitt): Import js types properly.
+import 'dart:typed_data';
+import 'package:js/js.dart';
```
srujzs, line 25: I think you're already tracking this, but eventually this should be dart:js_interop.

tool/bindings_generator/generate_bindings.mjs
```diff
@@ -0,0 +1,73 @@
...
+  const files = await idl.listAll();
+  const translator = new Translator();
+  for (const [shortname, file] of Object.entries(files)) {
+    translator.collect(await file.parse());
+  }
+  return translator.translate();
+}
+
+function runDartFormat(filePath) {
+  child_process.exec(`dart format ${filePath}`, (error, stdout, stderr) => {
```
srujzs, line 39: Nothing to address now, but I think we can avoid the exec call when this is migrated to Dart and use the dart formatter by source instead.

kevmoo: pkg:dart_style, yes!

PR #7, opened 2023-02-12
Port the majority of the bindings generator to Dart

joshualitt:


joshualitt:
@srujzs No rush on this, it can't land until there is a dev channel release of the Dart SDK.

tool/bindings_generator/dart_main.dart
```diff
@@ -0,0 +1,44 @@
...
+// Generates DOM bindings for Dart.
+// TODO(joshualitt): This prototype has exposed a limitation of JS types on JS
+// backends.
+//   * Static interop doesn't seem to work on some types, atleast JSArrays. We
+//     should fix this.
+//  * We need to have a more specific type for JSPromise.
+// TODO(joshualitt): Find a way to generate bindings for JS builtins. This will
+// probably involve parsing the TC39 spec.
+
+void runDartFormat(String dir) {
```
joshualitt, line 20: FYI @srujzs `DartFormat` gave me some issues when I ran it on Node, something with `Uri.base` or something. This may be fixable, but its easy to just call out to the host too.

joshualitt:
Woohoo, this PR is good to go. ptal when you have  a chance @srujzs

lib/dom.dart
```diff
@@ -2,10 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// TODO(joshualitt): Import js types properly.
-import 'dart:typed_data';
+import 'dart:js_interop';
```
srujzs, line 5: I'm genuinely surprised this lets you do this. It seems like Dart allows you to shadow exports? They have two `@JS` annotations and they're separate semantics (the `dart:js_interop` does type checks), so I think you want to `hide` the `package:js` one here and everywhere else.

I need to think about this a bit more because I was hoping there would be a static error when you import both, not whatever this is.

tool/bindings_generator/dart_main.dart
srujzs: Weird, but okay. I do want us to move to source at some point, so mind adding a TODO? It'll be useful to mention what issues you came across for the future.

srujzs (changes requested):

tool/bindings_generator/dart_main.dart
joshualitt: Whoops, good thing I looked a bit into this. We needed `self.location` set in our preamble.

srujzs (approved):

PR #8, opened 2023-02-12
Breakout `dom.dart` into multiple files

### LINKED ISSUE #3: Break out interop classes into multiple files.

joshualitt:
Fixes #3


tool/bindings_generator/generate_bindings.dart
```diff
@@ -8,22 +8,21 @@ import 'translator.dart';
...
 
+/// Returns the part of the header common to every Dart file generated from the
+/// IDL.
 String generateHeader() {
-  return '''// Copyright (c) $year, the Dart project authors.  Please see the AUTHORS file
-// for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICENSE file.
+  return '''$licenseHeader
 
+// ignore_for_file: unused_import
```
joshualitt, line 16: This is unfortunate, but to do better we'd have to do fine grained dependency tracking because the IDL seems to assume a global namespace. I'm willing to do this work, but it doesn't seem very valuable.

kevmoo: Maybe I'll do it in my free time 😄

kevmoo (approved):

kevmoo:
Can you remove the `// ignore` thingy now in the generated code?

joshualitt:
Unfortunately not quite yet. I left a TODO in the code, but basically the issue is the dependency on `package:js` for the `@staticInterop` annotation. I need to do a `hide` of `JS` from `package:js` , but I don't know a priori if we need `staticInterop`. This will go away in the near future, so we can clean this up.

But yea, code_builder is awesome!

tool/bindings_generator/generate_bindings.dart
joshualitt: This is done now, except for `package:js/js.dart hide JS`, there is a TODO in the code.

PR #20, opened 2023-02-20
Add support for dictionary constructors and fields

joshualitt:


tool/bindings_generator/translator.dart
```diff
@@ -312,6 +314,55 @@ class Translator {
...
+    }
+    switch (type) {
+      case 'boolean':
+        return (defaultValue as JSBoolean).toDart.toString();
+      case 'string':
+        return "'${(defaultValue as JSString).toDart}'";
+      case 'number':
+        return (defaultValue as JSString).toDart.toString();
+      case 'sequence':
+        // For sequenec, the only possible value is `[]`.
```
srujzs, line 334: nit: spelling

tool/bindings_generator/translator.dart
```diff
@@ -326,13 +377,14 @@ class Translator {
   }
 
   List<code.Expression> _jsOverride(String? jsOverride,
-          [bool staticInterop = false]) =>
+          {bool staticInterop = false, bool anonymous = false}) =>
```
srujzs, line 379: nit: We should keep the naming scheme the same e.g. objectLiteral instead of anonymous

lib/src/dom/accelerometer.dart
```diff
@@ -30,11 +30,16 @@ extension AccelerometerExtension on Accelerometer {
 
 @JS('AccelerometerSensorOptions')
```
srujzs, line 31: nit: You don't have to address this here, but we don't need this renaming for two reasons:

1. It's the same as the class name.
2. It's an @anonymous class so you don't even have a named constructor anymore.

lib/src/dom/accelerometer.dart
```diff
@@ -30,11 +30,16 @@ extension AccelerometerExtension on Accelerometer {
...
 @staticInterop
+@anonymous
 class AccelerometerSensorOptions extends SensorOptions {
-  external factory AccelerometerSensorOptions();
+  external factory AccelerometerSensorOptions(
+      {AccelerometerLocalCoordinateSystem referenceFrame = 'device'});
```
srujzs, line 36: As we know, this will be ignored with existing object literal semantics. In the future too, this will just make the default value explicit in the options object users will pass. As such, do you think there is value in having these defaults here?

srujzs (approved):

lib/src/dom/accelerometer.dart
joshualitt: Let's leave them now for documentation. It's true they aren't really used, but that doesn't matter in this case because they are 'default' from the perspective of the IDL. If we ever support default values, it won't be harmful to pass these to the IDL, but we might want to make them comments at that point just because it could possibly be a bit more efficient.

PR #25, opened 2023-02-22
no empty

kevmoo:
- tool: add CLI flag to delete src directory before generation
- Don't generate empty libraries


kevmoo:
see https://github.com/dart-lang/web/issues/26 instead

PR #27, opened 2023-03-01
tool: add CLI flag to delete src directory before generation

kevmoo:


tool/update_bindings.dart
```diff
@@ -44,6 +44,14 @@ $_usage''');
   }
 
   // Run app with `node`.
+  if (argResult['delete-src'] as bool) {
+    // TODO: file cleanup should happen in the tool itself
```
joshualitt, line 48: name or bug # in TODO?

tool/bindings_generator/dart_main.dart
```diff
@@ -20,6 +20,7 @@ void main(List<String> args) async {
   await _generateAndWriteBindings(args[0]);
 }
 
+// TODO: delete any file that hasn't been generated!
```
joshualitt, line 23: Maybe we should change this to say something like `After we break apart monolithic interfaces, assert we do not generate empty files`

joshualitt (approved):

PR #29, opened 2023-03-01
Cleanup handling of overridden functions / constructors

### LINKED ISSUE #9: Cleanup method / factory overriding.

joshualitt:
* Adds support for optional parameters, and uses them to support overridden functions / constructors.
* Changes all uses of `extends` to `implements`. This is to avoid having to generate empty factory constructors.

Fixes #9

### DIFF
--- a/tool/bindings_generator/translator.dart
+++ b/tool/bindings_generator/translator.dart
@@ -80,15 +80,145 @@ class _Library {
   }
 }
 
+String _typeRaw(idl.IDLType idlType) {
+  final String type;
+  if (idlType.union.toDart) {
+    type = 'union';
+  } else if (idlType.generic.toDart.isNotEmpty) {
+    type = idlType.generic.toDart;
+  } else {
+    type = (idlType.idlType as JSString).toDart;
+  }
+  if (typeAliases.containsKey(type)) {
+    return typeAliases[type]!;
+  } else {
+    return type;
+  }
+}
+
+class _Type {
+  String type;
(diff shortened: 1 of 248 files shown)


joshualitt:
@kevmoo, we can start the review now, but this can't land until https://dart-review.googlesource.com/c/sdk/+/285701 lands.

kevmoo:
@joshualitt – run the formatter quick!

tool/bindings_generator/translator.dart
```diff
@@ -80,15 +80,145 @@ class _Library {
...
+      type = 'JSAny';
+    }
+    if (idlType.nullable.toDart) {
+      isNullable = true;
+    }
+  }
+}
+
+class _Parameter {
+  Set<String> _names;
```
kevmoo, line 122: could be final?

joshualitt:
Lol, well, this passes but the behavior will be strange without the specialization CL, i.e. we'll pass `null` many places instead of nothing.

kevmoo (approved):
Did you consider updating the test file to use of these updated APIs?

joshualitt:
> Did you consider updating the test file to use of these updated APIs?

I think we have to walk a fine line with testing in this package. We definitely need tests of the generator, I created #30 to track that effort. But I don't think we want to test the semantics of JS interop here, we should do that in the SDK. Nor do we want to test browser conformance.

That said, I'm not against more smoke tests. Let's think this through in #30.

Issue #26, opened 2023-03-01
Break apart monolithic interfaces


joshualitt:
Today we implement partial interfaces by just collecting everything into a single interface. This results in empty files, among other issues. We should fix this by breaking apart monolithic interfaces, and putting the partial interfaces alongside the WebIDL files that implement them.

PR #33, opened 2023-03-06
Don't generate empty libraries

kevmoo:


joshualitt:
Wait, I thought we wanted to fix this a different way? I agree we shouldn't be generating empty libraries, but I think the fix is to put something in them?

kevmoo:
Once we get the fix in that eliminates empty libraries, this will be a no-op.

Honestly, I think it's a nice signal.

```
Skipped empty files:
  src/dom/beacon.dart
  src/dom/compat.dart
  src/dom/csp_embedded_enforcement.dart
  src/dom/css_images_4.dart
  src/dom/css_nesting.dart
  src/dom/css_shadow_parts.dart
  src/dom/html_media_capture.dart
  src/dom/input_events.dart
  src/dom/permissions_request.dart
  src/dom/permissions_revoke.dart
  src/dom/pointerlock.dart
  src/dom/prefer_current_tab.dart
  src/dom/prerendering_revamped.dart
  src/dom/priority_hints.dart
  src/dom/private_click_measurement.dart
  src/dom/storage_access.dart
  src/dom/wasm_web_api.dart
  src/dom/webrtc_svc.dart
  src/dom/webxr_gamepads_module.dart
```

Might be nice to see "oh, there's nothing generated there" 🤷

joshualitt:
If it's okay with you, can we first investigate whether or not it is possible to actually put code in these libraries? If it isn't, then I'm happy to take this CL. If it is, then that is the right way to fix this issue. Yes, if we put code in these files, then this logic becomes a noop, but it also just masks a design problem. I think ideally we'd have an assert that we don't generate any files, because it could hide real bugs.

Thoughts?

kevmoo:
@joshualitt – you're the engineer here. 😄 

If we just created separate extension types in each file now, that's great.

You can't really split `inline class` across files, though. I think we'll need *one*

joshualitt:
Yea, I was thinking about the inline class case too, and maybe that means we do need this logic.

On the other hand, we can continue to use extensions if we want to have separation here, and maybe some day we'll have inline class mixins.

Taking a step back, RE: inline classes, we know we want to use them for types / constructors, but do we also want to use them for methods? One reason to continue to prefer a separation between types and their methods is to make it easier for users to 'override' the behavior of these types depending on their needs. For example, if they want 'dartier' style methods, vs. low level JS methods, it is easier for them to do this if we keep the types separate from their methods. Of course the 'downside' to this approach is they have to import `package:web` every time they want to use methods on these objects, but TBH I'm not sure this is really a downside.

Anyway, I haven't really thought this through yet, so I'm just thinking out loud, but it is food for thought.

Sorry to belabor this, my point is lets leave this CL up for now. These empty files aren't hurting anyone. The separately, let's have a discussion about how `package:web` will be structured with inline classes. If we intend to have monolithic interfaces, then I think that is a strong case to land this CL now. If we intend to always have some separation between `package:web` types and their methods, then let's break apart these monolithic interfaces now, and fix the empty files that way.

I created #26 to track this.

PR #34, opened 2023-03-27
Remove default values.

joshualitt:


joshualitt:
@srujzs ptal

joshualitt:
Well for one, we ignore them, and if we didn't we'd have inconsistent semantics between tearoffs and static invocations.

But for another, for the same reason default values don't mean anything on abstract declarations. The 'defaultness' is implemented by the target of these externals, not by our interop layer.

kevmoo:
> Well for one, we ignore them, and if we didn't we'd have inconsistent semantics between tearoffs and static invocations.
> 
> But for another, for the same reason default values don't mean anything on abstract declarations. The 'defaultness' is implemented by the target of these externals, not by our interop layer.

even as documentation they are SUPER useful, I think.
Is there any downside to having them? (Other than potential conflicts with what the runtime system implements and what's in the IDL)

joshualitt:
> even as documentation they are SUPER useful, I think. Is there any downside to having them? (Other than potential conflicts with what the runtime system implements and what's in the IDL)

So far we've been saying that documentation for the DOM belongs outside of `package:web` i.e. in the MDN, so while I understand it's kind of nice to have this, removing them is consistent with our policy so far for the rest of `package:web`.

But really I'm removing this because it blocks rollout of `package:web` on Dart2Wasm. Specifically, we don't yet have implicit conversions for Dart types to JS types, so `package:web` currently fails to compile on Dart2Wasm. This CL fixes that issue. We can always revisit this after we have implicit conversions, but TBH I think we should have a general discussion on documenting `package:web` vs. relying on external documentation sources before we do so.

kevmoo:
Okay. Just wanted to check. Thanks for the details!

srujzs:
> Well for one, we ignore them, and if we didn't we'd have inconsistent semantics between tearoffs and static invocations.

Note that until we disallow tear-offs, we still have inconsistent semantics. The tear-off will use the default value while a direct call will not.

Another thing to note is that if the default value of a browser API changes, and our package has not yet, this is confusing for users who see a default value in the package. It never gets passed and the browser uses its own value.

In general, default values should be disallowed on external methods if we never use them. If we want documentation, I prefer we use comments to indicate what the default will be.

kevmoo:
@srujzs – this should be flagged by the analyzer then! At least a warning, no?

srujzs:
> @srujzs – this should be flagged by the analyzer then! At least a warning, no?

The plan is to indeed disallow them through the CFE (js_interop_checks) for newer syntax (inline classes).

Eventually it's worthwhile to port all our interop checks to analyzer so people get errors sooner, but that's part of a larger future effort. :)

PR #40, opened 2023-04-24
Remove dependency on `package:js`.

joshualitt:


srujzs:
LGTM, but the pubspec needs to be updated so it can use https://github.com/dart-lang/sdk/commit/568469b8967dfc6341f502211b65ff8cfe056c9b (with min constraint >= 3.1.0-0.0.dev).

PR #48, opened 2023-05-19
Add support for accessing CSS properties directly from JS.

### LINKED ISSUE #47: Add helpers for common css properties
In dart:html we generated many common CSS properties as properties, so you wouldn't have to call `setProperty` by hand. Browsers anyways accept setting these directly by assigning the underlying property on the css declaration object.

See for example: https://github.com/dart-lang/sdk/blob/main/tools/dom/scripts/CSSPropertyNames.in

joshualitt:
Fixes #47


joshualitt:
@srujzs @sigmundch ptal. I'm not sure if we really want to expose the webkit properties in this manner. I'm happy to hide those if we don't think they are useful.

kevmoo:
Cool!

I'd probably default to dropping the browser-specific ones. Unless we get individual requests.

tool/bindings_generator/generate_bindings.dart
```diff
@@ -7,11 +7,48 @@ import 'dart:js_util';
...
 import 'webref_idl_api.dart';
 
+/// Generate CSS property names for setting / getting CSS properties in JS.
+Future<List<String>> _generateCSSStyleDeclarations() async {
+  final cssStyleDeclarations = <String>{};
+  final array = objectEntries(await promiseToFuture<JSObject>(css.listAll()));
+  for (var i = 0; i < array.length.toDart; i++) {
+    final entry = array[i.toJS] as JSArray;
+    final data = entry[1.toJS] as JSObject;
+    final properties = getProperty<JSArray?>(data, 'properties');
```
sigmundch, line 20: any reason why not add a couple static-interop classes to represent the array entries and property entries?

tool/bindings_generator/generate_bindings.dart
```diff
@@ -7,11 +7,48 @@ import 'dart:js_util';
...
+import 'webref_css_api.dart';
 import 'webref_idl_api.dart';
 
+/// Generate CSS property names for setting / getting CSS properties in JS.
+Future<List<String>> _generateCSSStyleDeclarations() async {
+  final cssStyleDeclarations = <String>{};
+  final array = objectEntries(await promiseToFuture<JSObject>(css.listAll()));
+  for (var i = 0; i < array.length.toDart; i++) {
+    final entry = array[i.toJS] as JSArray;
+    final data = entry[1.toJS] as JSObject;
```
sigmundch, line 19: not necessarily actionable right now: all these index accesses feel very cumbersome, I think we will want to continue iterating on how to improve the experience here.

My default at the moment would be to create a proxy to wrap the array, use regular iteration, and use a Dart int indices.

joshualitt:
> Cool!
> 
> I'd probably default to dropping the browser-specific ones. Unless we get individual requests.

done.

tool/bindings_generator/generate_bindings.dart
```diff
@@ -7,11 +7,50 @@ import 'dart:js_util';
...
+    final properties = getProperty<JSArray?>(data, 'properties');
+    if (properties != null) {
+      for (var j = 0; j < properties.length.toDart; j++) {
+        final property = properties[j.toJS] as JSObject;
+        // There are thre cases for [styleDeclaration]:
+        //   1) Length == 1, a single word CSS property.
+        //   2) Length == 2, a keybab case property + a camel case property.
```
sigmundch, line 26: what is a `keybab`?

sigmundch (approved):

tool/bindings_generator/generate_bindings.dart
joshualitt: whoops, typo. Fixed to be 'kebab'.

tool/bindings_generator/generate_bindings.dart
joshualitt: Yea, we definitely want to do something better here.

As you suggest, for places where performance isn't a concern, then an adapter class seems like a good choice.

We can provide such helper adapters, perhaps we can even avoid reifying the adapters on JS backends.

PR #52, opened 2023-06-12
Use Dart primitive types for JS primitives.

joshualitt:


pubspec.yaml
kevmoo, line 1: Do we need to bump the min supported SDK?

tool/bindings_generator/translator.dart
```diff
@@ -370,9 +370,15 @@ class Translator {
...
     // Unfortunately, `code_builder` doesn't know the url of the library we are
     // emitting, so we have to remove it here to avoid importing ourselves.
-    // TODO(joshualitt): Properly track JS type dependencies.
-    String? url = _typeToLibrary[symbol]?.url ?? 'dart:js_interop';
-    if (url == _currentlyTranslatingUrl) {
+    var url = _typeToLibrary[symbol]?.url;
+
+    // JS types and core types don't have urls.
+    if (url == null) {
+      if (symbol.startsWith('JS')) {
```
srujzs, line 377: This may be insufficient if we decide to name the JS types differently. Why is the previous code insufficient here now?

srujzs (approved):

tool/bindings_generator/translator.dart
srujzs: We're not renaming, I'm just worried that this code is brittle because it relies on our naming scheme.

Since we don't have any other information besides the name though, I don't think we can write this better.

PR #59, opened 2023-06-29
[web] introduce the helper layer.

sigmundch:
This change introduces the helper layer. It was built by hand and it is not meant to be comprehensive or complete.

Currently we include:
- implementation of event streams
- many common events added on `Element` (this was a large list copied from `dart:html`)
- a few extensions to introduce some renames discovered by hand while migrating one application.
- a couple top-level helper methods often used in `dart:html`
- the highly used HttpRequest static methods
- enums exposed in `dart:html`
- a conversion for TouchList
- some type aliases (currently hand curated, we should expand the list of renames in a more automated way)

_Edit: renamed glue layer to helper layer._

### DIFF
--- a/lib/src/helpers/enums.dart
+++ b/lib/src/helpers/enums.dart
@@ -7,7 +7,7 @@ library;
 
 import '../../web.dart';
 
-class KeyLocation {
+abstract final class KeyLocation {
   // Note: this library uses legacy constant names for backwards compatibility.
   // ignore: constant_identifier_names
   static const int STANDARD = 0;
@@ -23,7 +23,7 @@ class KeyLocation {
   static const int MOBILE = 5;
 }
 
-abstract class KeyCode {
+abstract final class KeyCode {
   // These constant names were borrowed from Closure's Keycode enumeration
   // class.
   // https://github.com/google/closure-library/blob/master/closure/goog/events/keycodes.js
@@ -214,7 +214,7 @@ abstract class KeyCode {
   static const int NUM_PERIOD = 110;
 }
 
-class Device {
+abstract final class Device {
   static String get userAgent => window.navigator.userAgent;
   static final bool isOpera = userAgent.contains('Opera');
   static final bool isIE = !isOpera && userAgent.contains('Trident/');
--- a/lib/src/helpers/enums.dart
+++ b/lib/src/helpers/enums.dart
@@ -2,8 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// Glue layer library that exposes enums commonly used from `dart:html`.
-library web.glue.enums;
+/// Helper layer library that exposes enums commonly used from `dart:html`.
+library;
 
 import '../../web.dart';
 
--- a/lib/src/helpers/enums.dart
+++ b/lib/src/helpers/enums.dart
@@ -5,7 +5,7 @@
 /// Glue layer library that exposes enums commonly used from `dart:html`.
 library web.glue.enums;
 
-import '../web.dart';
+import '../../web.dart';
 
 class KeyLocation {
   // Note: this library uses legacy constant names for backwards compatibility.
--- a/lib/src/helpers/events/providers.dart
+++ b/lib/src/helpers/events/providers.dart
@@ -5,8 +5,8 @@
 import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
 
-import '../../web.dart';
-import '../glue.dart' show Device;
+import '../../../helpers.dart' show Device;
+import '../../../web.dart';
 import 'streams.dart';
 
 /// Statically accessible `EventStreamProvider`s for all event types.
--- a/lib/src/helpers/http.dart
+++ b/lib/src/helpers/http.dart
@@ -6,7 +6,7 @@ import 'dart:async';
 import 'dart:js_interop';
 import 'dart:typed_data';
 
-import '../web.dart';
+import '../../web.dart';
 import 'events/events.dart' show XHRGetters;
 
 /// A helper used to make it easier to operate over [XMLHttpRequest]s.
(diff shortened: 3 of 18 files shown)


joshualitt (approved):
This looks good. I think the one thing we might want to bikeshed briefly on is the name.

I'm okay with `glue` for a v0, but it might be nice to pick a name that more closely reflects the purpose of this library. Some options:
`dart_html`
`darty_web` / `darty_dom`
`legacy` / `migration` / `transition` perhaps as a prefix with `_html` or `_web`
`helper` / `util` / `ext`

That said, I'm okay with `glue` as long as we document the meaning in the readme.

sigmundch:
Thanks - happy to change the name, let's brainstorm some more. When chatting yesterday with Srujan, we came across the general question on how static_html will share some of the logic with dart:html, and potentially that we could make the streams and other helpers also be shared with package:web. So those considerations may affect what name we want to pick here.

I made an extra, but small documentation change to highlight one difference between the streams here and the ones in dart:html

srujzs:
LGTM for an initial commit - we may revisit later and pare things down or add more helpers later as people make requests.

sigmundch:
@kevmoo - PTAL, in particular:
* this is all updated renaming "glue layer" to "helper layer"
* I still reexport web.dart from helpers.dart - do you prefer we don't? It just makes it easier to only have to do one vs two imports.

The failures on the CI seem unrelated to my change, they seem to come from changes in the SDK to remove `.toDart` (and replace it with `.toDartInt`). We should do a separate PR to fix that.

lib/src/helpers/enums.dart
```diff
@@ -0,0 +1,226 @@
...
+  static const int NUM_MULTIPLY = 106;
+  // ignore: constant_identifier_names
+  static const int NUM_PLUS = 107;
+  // ignore: constant_identifier_names
+  static const int NUM_MINUS = 109;
+  // ignore: constant_identifier_names
+  static const int NUM_PERIOD = 110;
+}
+
+class Device {
```
kevmoo, line 217:
```suggestion
abstract final class Device {
```

lib/src/helpers/enums.dart
```diff
@@ -0,0 +1,226 @@
...
+  static const int RIGHT = 2;
+  // ignore: constant_identifier_names
+  static const int NUMPAD = 3;
+  // ignore: constant_identifier_names
+  static const int JOYSTICK = 4;
+  // ignore: constant_identifier_names
+  static const int MOBILE = 5;
+}
+
+abstract class KeyCode {
```
kevmoo, line 26:
```suggestion
abstract final class KeyCode {
```

lib/src/helpers/enums.dart
```diff
@@ -0,0 +1,226 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Helper layer library that exposes enums commonly used from `dart:html`.
+library;
+
+import '../../web.dart';
+
+class KeyLocation {
```
kevmoo, line 10:
```suggestion
abstract final class KeyLocation {
```

lib/src/helpers/events/providers.dart
```diff
@@ -0,0 +1,627 @@
...
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+
+import '../../../helpers.dart' show Device;
+import '../../../web.dart';
+import 'streams.dart';
+
+/// Statically accessible `EventStreamProvider`s for all event types.
+class EventStreamProviders {
+  static const EventStreamProvider<ProgressEvent> abortEvent =
```
kevmoo, line 14: please tell me we have a script for this!

lib/src/helpers/http.dart
```diff
@@ -0,0 +1,251 @@
...
+///
+/// ## Other resources
+///
+/// * [Fetch data dynamically](https://dart.dev/tutorials/web/fetch-data/),
+/// a tutorial shows how to load data from a static file or from a server.
+/// * [Dart article on using HttpRequests](https://dart.dev/guides/libraries/library-tour#using-http-resources-with-httprequest)
+/// * [JS XMLHttpRequest](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest)
+/// * [Using XMLHttpRequest](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest/Using_XMLHttpRequest)
+
+class HttpRequest {
```
kevmoo, line 51: Should have a TODO Here to provide a `pkg:http` wrapper!

kevmoo (approved):

lib/src/helpers/http.dart
```diff
@@ -0,0 +1,251 @@
...
+/// app that uses it.
+///
+/// ## Other resources
+///
+/// * [Fetch data dynamically](https://dart.dev/tutorials/web/fetch-data/),
+/// a tutorial shows how to load data from a static file or from a server.
+/// * [Dart article on using HttpRequests](https://dart.dev/guides/libraries/library-tour#using-http-resources-with-httprequest)
+/// * [JS XMLHttpRequest](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest)
+/// * [Using XMLHttpRequest](https://developer.mozilla.org/en-US/docs/DOM/XMLHttpRequest/Using_XMLHttpRequest)
+
```
kevmoo, line 50: delete newline?

lib/src/helpers/events/providers.dart
sigmundch: nope - copied by hand from dart:html. There was a generator there, but we are not carrying that over.

lib/src/helpers/http.dart
sigmundch: I don't follow - shouldn't that be in pkg:http instead? (replace their dart:html import by a package:web import?)

kevmoo: Eventually. I guess we should keep this simple. We should do that migration as soon as we can, though...

lib/src/helpers/enums.dart
sigmundch: done

lib/src/helpers/enums.dart
sigmundch: done

lib/src/helpers/enums.dart
sigmundch: done

lib/src/helpers/http.dart
sigmundch: done

PR #68, opened 2023-09-18
Fix globalContext rename

kevmoo:


lib/helpers.dart
```diff
@@ -53,7 +53,7 @@ Element? querySelector(String selectors) => document.querySelector(selectors);
 
 bool isInstanceOfDomType(JSObject? o, String domType) {
   if (o == null) return false;
-  final constructor = globalJSObject[domType.toJS];
+  final constructor = globalContext.getProperty(domType.toJS);
```
sigmundch, line 56: I believe you can do `globalContext[domType]`  (now `[]` takes Strings instead of JSStrings to make it easier to write).

pubspec.yaml
```diff
@@ -6,7 +6,7 @@ version: 0.1.4-beta
 repository: https://github.com/dart-lang/web
 
 environment:
-  sdk: ">=3.1.0-185.0.dev <4.0.0"
+  sdk: ">=3.2.0-75.0.dev <4.0.0"
```
sigmundch, line 9: However the String-based `[]` came later than this version. Looking at https://github.com/dart-lang/sdk/commit/4efb33e1281, I believe you need `3.2.0-157.0.dev` instead

sigmundch (approved):
Seems there is an info error that is making the bots unhappy, should we fix that too?

After those, we should bump and publish the package. Not sure if we have set up the auto-publish bot on this repo though.

PR #73, opened 2023-09-19
helpers nit: mark classes with only statics `sealed`

kevmoo:
Fix some relative imports

### DIFF
--- a/lib/src/helpers/enums.dart
+++ b/lib/src/helpers/enums.dart
@@ -5,7 +5,7 @@
 /// Helper layer library that exposes enums commonly used from `dart:html`.
 library;
 
-import '../../web.dart';
+import '../dom/html.dart';
 
 abstract final class KeyLocation {
   // Note: this library uses legacy constant names for backwards compatibility.
--- a/lib/src/helpers/events/providers.dart
+++ b/lib/src/helpers/events/providers.dart
@@ -5,12 +5,12 @@
 import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
 
-import '../../../helpers.dart' show Device;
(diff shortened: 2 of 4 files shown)


lib/src/helpers/enums.dart
```diff
@@ -5,9 +5,9 @@
 /// Helper layer library that exposes enums commonly used from `dart:html`.
 library;
 
-import '../../web.dart';
+import '../dom/html.dart';
 
-abstract final class KeyLocation {
```
srujzs, line 10: Why `sealed` over `abstract final`? These don't have constructors and we don't really need exhaustiveness (although I suppose `abstract sealed` doesn't hurt).

lib/src/helpers/events/providers.dart
```diff
@@ -5,12 +5,12 @@
...
 import 'dart:js_interop_unsafe';
 
-import '../../../helpers.dart' show Device;
 import '../../../web.dart';
+import '../enums.dart' show Device;
 import 'streams.dart';
 
 /// Statically accessible `EventStreamProvider`s for all event types.
-class EventStreamProviders {
+sealed class EventStreamProviders {
```
srujzs, line 13: Could also make this `abstract sealed`.

kevmoo: `abstract sealed` just adds the ability to extend...which...why?

https://dart.dev/language/modifier-reference

srujzs: Turns out `abstract sealed` doesn't even exist. :) `abstract final` it is then.

srujzs (approved):

PR #75, opened 2023-09-19
Expose EventStreamProvider from helpers, prepare 0.2.1-beta release

kevmoo:


srujzs:
I think this is fine to expose. Even if we exposed all the `EventStreamProvider`s that `dart:html` exposes, we'd still be missing a few events that have been added since the last IDL roll e.g. `beforexrselect`.

srujzs (approved):

Issue #57, opened 2023-06-27
Consider using `num` instead of `double` for parameter types


sigmundch:
This is more backwards compatible with types used in dart:html, and hence less work to adapt a call that used to pass num or int arguments.

srujzs:
Reopening due to concerns around users accidentally downcasting `num` in dart2wasm because they may assume dart2wasm has the same number semantics as the JS compilers (it does not).

A compromise here to:

- Make migration and passing in integer values easier
- Avoid the accidental downcasting

is to keep the parameter types as `num` but the return types as `double`.

cc @eyebrowsoffire

parlough:
I also think this would be worth reevaluating. My feeling is that today's Dart developers aren't as familiar with `num` as they used to be.

Cases passing int literals would keep working with `double` and since there are changes you need to make when switching to `package:web` anyway, I think handling the occasional number change shouldn't be too bad. It might even be a helpful thing for that decision to be made more consciously.

srujzs:
I agree. I want to perhaps include this in the next major release (which we'll certainly have as we tackle other issues). It does mean a little bit of pain for migration, but it should be minimal.

> My feeling is that today's Dart developers aren't as familiar with num as they used to be.

Most developers aren't aware of the number differences partly because it derives from a web vs native difference. Now that we have the number difference within web, this will pop up more often.

PR #78, opened 2023-09-25
Change parameter double types to num

srujzs:
Closes https://github.com/dart-lang/web/issues/57

Makes it easier to migrate off of dart:html APIs and allows users to pass integers without having to convert to doubles.


srujzs:
We could go further and replace all `double`s with `num` for consistency (and we wouldn't have to worry about typedefs), but that's a breaking change. Thoughts?

srujzs:
Chatted offline, but summarizing: it's quite rare for `dart:html` to return a `double`, while it's much more common for it to use `num`s. As such, it's fine to use `num` for consistency and not have to worry about typedefs.

sigmundch (approved):

Issue #77, opened 2023-09-21
Refactor use of typeofEquals and instanceof and remove isInstanceOfDomType


srujzs:
`dart:js_interop` will have the following changes with https://dart-review.googlesource.com/c/sdk/+/326691:

 - `typeofEquals` takes a `String` and returns `bool`
 - `instanceof` returns a `bool`
 - `instanceOfString` is added to check `instanceof` with types that are available in the global context. `isInstanceOfDomType` does the same thing and can be removed.

This will require publishing a new version and updating the minimum SDK.

kevmoo:
Are the changes to `dart:js_interop` breaking? Do we have the changelog updated?

srujzs:
Yes, and yes in the above CL once it lands. It makes sense to have these go in before 3.2 releases before more packages migrate.

PR #80, opened 2023-09-25
Refactor typeofEquals usage and remove isInstanceOfDomType

srujzs:
Closes https://github.com/dart-lang/web/issues/77

typeofEquals now takes a String and returns a bool, and instanceOfString is the canonical replacement for isInstanceOfDomType. Removes now unneeded test as well. Updates min SDK version.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,12 +1,12 @@
 name: web
 description: >-
   Lightweight DOM and JS bindings built around JS static interop.
-version: 0.2.1-beta
+version: 0.2.2-beta
 
 repository: https://github.com/dart-lang/web
 
 environment:
-  sdk: ">=3.2.0-157.0.dev <4.0.0"
+  sdk: ">=3.2.0-194.0.dev <4.0.0"
 
 dev_dependencies:
   args: ^2.4.0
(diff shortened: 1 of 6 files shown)


srujzs:
This will have to wait until we publish a Dart dev version >= this minimum.

pubspec.yaml
```diff
@@ -6,7 +6,7 @@ version: 0.2.1-beta
 repository: https://github.com/dart-lang/web
 
 environment:
-  sdk: ">=3.2.0-157.0.dev <4.0.0"
+  sdk: ">=3.2.0-194.0.dev <4.0.0"
```
kevmoo, line 9: double check pubspec version and changelog version

srujzs: Done, thanks!

sigmundch (approved):
Overall looks good, but I'd also be OK keeping a test like the one deleted in helpers_test. That's because, while technically the `instnaceOfString` should be tested by the SDK, there is still value in having a test here that ensures that it works for package:web types.

srujzs:
> Overall looks good, but I'd also be OK keeping a test like the one deleted in helpers_test. That's because, while technically the `instnaceOfString` should be tested by the SDK, there is still value in having a test here that ensures that it works for package:web types.

Fair point, added it back and refactored to account for `instanceOfString`.

sigmundch (approved):
Thanks! :+1:

Issue #65, opened 2023-08-02
Expose types of `TypedData` or `JSTypedArray`


kevmoo:
```
typedef ArrayBufferView = JSAny?;
typedef BufferSource = JSAny?;
typedef AllowSharedBufferSource = JSAny?;
```

This may be a more specific version of an issue related to

https://github.com/dart-lang/web/blob/0344abd564f2094fb476f3d88cfbf1a581eb6e03/tool/bindings_generator/translator.dart#L110-L113

kevmoo:
See https://webidl.spec.whatwg.org/#ArrayBufferView and https://webidl.spec.whatwg.org/#BufferSource

srujzs:
Note that the first two typedefs will resolve to `JSObject` since they are a union of `DataView`, which doesn't implement `TypedArray`. The last one is a union of `SharedArrrayBuffer` which doesn't have a supertype, although maybe we can include it as an alias in our type_aliases? We default to `JSAny?` in a lot of cases but maybe we can and should do better and default to `JSObject` instead.

srujzs:
Union types have been added.

PR #86, opened 2023-10-14
Use least common supertypes for unions

srujzs:
https://github.com/dart-lang/web/issues/65

It's beneficial to have a more restrictive type for some APIs to get better typing.

1. Precomputes a map of least common supertypes of JS types
2. Modifies type mapping to handle union types
3. Desugars typedefs in the case of union types


tool/bindings_generator/js_types_least_common_supertypes.dart
```diff
@@ -0,0 +1,632 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Updated by /tool/update_bindings.dart. Do not modify by hand.
+
+const Map<String, Map<String, Set<String>>> jsTypesLeastCommonSupertypes = {
```
sigmundch, line 7: I'm not sure I have the full context, but I am wondering whether we can use a different model for computing the least upperbound that doesn't require generating the expansion of the LUB for each pair of types here.

I believe the JS types form a simple tree-like structure in their subtyping relation, so I am hoping if we use that tree structure instead, it may simplifies the generator and the rules here.

srujzs: There isn't much more context than what's here. :)

A few comments:
- JS types are indeed a tree, but we may or may not change that in the future, but we could just move up the assertions for now and revisit if the type hierarchy gets complex. I'll add that revision.
- We can definitely just output a type hierarchy, and that's what I initially did, but it slows down the generator. You're going to be computing a LUB every time you come across a union rather than looking at a cache. Practically, it might not make much of a difference in terms of performance. I'll add the revision and see what it looks like.

sigmundch: Thanks! Regarding performance: We could dynamically cache results too, that could potentially be a middle ground to get something that is more compact but still efficient.

srujzs: Done, PTAL. Some comments on the diff:

- I don't think we need to cache results, the hierarchy is quite flat, the LUB calculation is quick, and there aren't *that* many unions.
- I had to add some additional changes to make this work by moving the updates of partial interfaces to another phase. For context, when we parse the AST and compute an intermediate representation, we resolve partial interfaces by computing unions of their type signatures. However, we haven't visited all the types yet, so this calculation might lead to incorrect results. So, this resolution of partial interfaces is moved to the step after we compute the intermediate representation, but before the translation/emission.
- I added some extra support for other types in the union calculation e.g. interfaces and updated some missing types in our type aliases.

Thanks!

tool/bindings_generator/translator.dart
```diff
@@ -80,19 +86,112 @@ class _Library {
...
+    t2Supertypes.add(type2);
+  }
+  var t1i = t1Supertypes.length - 1;
+  var t2i = t2Supertypes.length - 1;
+  while (t1i >= 0 && t2i >= 0) {
+    if (t1Supertypes[t1i] != t2Supertypes[t2i]) break;
+    t1i--;
+    t2i--;
+  }
+  return (t1Supertypes[t1i + 1], nullable);
```
sigmundch, line 154: I believe the logic is correct, but looking at this at first I was tempted to play through a few corner cases to feel more certain that this does what we intend it to do.

Not sure if it's worthwhile, but we could consider adding a small unit test to cover a couple corner cases (e.g. types near or at the root, equal types, strict subtypes, changing which type is t1 vs t2). I know equal types are handled early and don't hit this code path, but maybe we can pull out this part of the logic to test it separately (past where we computed the canonical type representation).

tool/bindings_generator/translator.dart
```diff
@@ -31,11 +33,14 @@ class _Library {
     final name = named.name.toDart;
     assert(!translator._typeToLibrary.containsKey(name));
     translator._typeToLibrary[named.name.toDart] = this;
+    translator._typeToDeclaration[named.name.toDart] = node;
```
sigmundch, line 36: (not for this CL): just realized webidl_api could be updated to start using primitive types directly. It will be a small but nice cleanup on the generator

sigmundch (approved):
Thanks @srujzs - looks great!

I added a couple small style suggestions below, but LGTM

tool/update_bindings.dart
```diff
@@ -131,6 +142,68 @@ Future<void> _runProc(
...
+            // TODO(srujzs): We'll likely need to change this once JS types move
+            // to extension types.
+            element.library.isInSdk && element.library.name == '_js_types';
+
+        if (!isInJsTypes(element)) return;
+        final supertypes = <String>{};
+        final immediateSupertypes = <InterfaceType>[
+          if (element.supertype != null) element.supertype!,
+          ...element.interfaces,
+        ];
```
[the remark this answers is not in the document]
srujzs: Filtered `Object` and added an assertion (which should have been <= 1). Changed the other assertion to ensure this is a tree.

tool/bindings_generator/translator.dart
```diff
@@ -80,19 +86,112 @@ class _Library {
...
+  final nullable = nullable1 || nullable2;
+  if (!jsTypeHierarchy.containsKey(type1) ||
+      !jsTypeHierarchy.containsKey(type2)) {
+    return ('JSAny', nullable);
+  }
+  // Compute path from root and find the last type that exists in both paths for
+  // a least upper bound.
+  final t1Supertypes = [type1];
+  final t2Supertypes = [type2];
+  while (jsTypeHierarchy[type1]!.isNotEmpty) {
```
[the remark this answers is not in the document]
srujzs: I chose them all. :)

It reads better now, thanks for the suggestions!

tool/bindings_generator/translator.dart
srujzs: I think that's fair. I've moved this code to a separate helper and added a test for it.

Re: equal types - equal "raw" types are handled early on. But if two different raw types map to the same *JS type*, we should handle that with this helper. I added a test for equal types to make sure.

A longer term goal for testing might be to get these scripts accepting a "fake" IDL for end-to-end testing, but that's probably a stretch goal. :)

tool/bindings_generator/translator.dart
srujzs: Yup, I'll make a note to address this in another CL.

sigmundch (approved):
Thanks again! LGTM!

test/type_union_test.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+import '../tool/bindings_generator/type_union.dart';
+
+void main() {
+  test('Unrelated types', () {
```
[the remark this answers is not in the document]
srujzs: Done, thanks!

PR #87, opened 2023-10-19
Remove unnecessary JS types from webidl_api

srujzs:
All of these go through .toDart conversions anyways, so we should just use the Dart primitive instead of the JS type primitive on these members.


sigmundch (approved):
Yay, thanks!

Issue #56, opened 2023-06-23
Generate static properties for `const` IDL entries


sigmundch:
It would be useful to expose all const properties from the IDL. For example, currently we are missing about ~300 GLenum values in `WebGLRenderingContext`, such as [`WebGLRenderingContext.DEPTH_BUFFER_BIT`][1]

[1]: https://registry.khronos.org/webgl/specs/latest/1.0/#5.14.11

sigmundch:
I just noticed that the current package exposes these values under `WebGLRenderingContextBase`. However, this type is not exposed by browsers, so they are not accessible with the current API (instead you get an error if you try to access them).

srujzs:
I think this is a bug in how we handle `interface mixin`s. While the solution of generating a Dart interface for these mixins and having interfaces that include those mixins `implement` the Dart interface works for instance members, it doesn't work for static members, as they are tied to the interface they are defined in.

It might be better to instead always add the members of the `interface mixin`s to the interfaces that include them and not generate an interface for the mixin at all.

kevmoo:
Just hit this with https://html.spec.whatwg.org/multipage/media.html#error-codes on 
https://github.com/flutter/packages/pull/7012#discussion_r1702435508

kevmoo:
Now that we're using extension types, could we do

```diff
diff --git a/lib/src/dom/html.dart b/lib/src/dom/html.dart
index 556e65f..475e7cb 100644
--- a/lib/src/dom/html.dart
+++ b/lib/src/dom/html.dart
@@ -3693,10 +3693,10 @@ extension type HTMLMediaElement._(JSObject _) implements HTMLElement, JSObject {
 /// API documentation sourced from
 /// [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/MediaError).
 extension type MediaError._(JSObject _) implements JSObject {
-  external static int get MEDIA_ERR_ABORTED;
-  external static int get MEDIA_ERR_NETWORK;
-  external static int get MEDIA_ERR_DECODE;
-  external static int get MEDIA_ERR_SRC_NOT_SUPPORTED;
+  static const int MEDIA_ERR_ABORTED = 1;
+  static const int MEDIA_ERR_NETWORK = 2;
+  static const int MEDIA_ERR_DECODE = 3;
+  static const int MEDIA_ERR_SRC_NOT_SUPPORTED = 4;
 
   /// The read-only property **`MediaError.code`** returns a numeric
[...]

PR #90, opened 2023-10-27
Don't emit interface mixins

srujzs:
They should be added to the interfaces they are mixed into. This allows for external static members to use the right nominal type. This CL also adds some logic to handle name collisions between instance and static operations, which is now needed for exactly one pair of operations: Response.prototype.json and Response.json. Also enables asserts in the translator.

https://github.com/dart-lang/web/issues/56


sigmundch (approved):
Nice!

kevmoo (commented):
Uh. Doesn't this mean, in some cases, you won't be able to polymorph over similar types since they won't share a base type?

srujzs:
That's true, but that's intentional, since they're not supposed to be used as types.

https://webidl.spec.whatwg.org/#idl-interface-mixins

> Note that unlike [interfaces](https://webidl.spec.whatwg.org/#dfn-interface) or [dictionaries](https://webidl.spec.whatwg.org/#dfn-dictionary), [interface mixins](https://webidl.spec.whatwg.org/#interface-mixin) do not create types.

Do you have an example where they are that I overlooked? `GenericTransformStream` isn't used as a type in the IDL afaict. If they are used as types, then one option is to just move the `static` members to the mixing-in class and keep the instance members and mixin interface around.

kevmoo:
> Do you have an example where they are that I overlooked?

I can imagine places where writing helpers/extensions on the shared types could be nice. I mean, we can always go back to these later. This is ABSOLUTELY a breaking change, though.

...I mean, if folks are actually using these types already. 🤷‍♂️

srujzs:
Searching GitHub gave me no results for these mixin types being used with `package:web`, but that isn't too surprising. I've added this change to the changelog, and if we change our minds later about wanting these types to be emitted, we can go the route of [comment#5.](https://github.com/dart-lang/web/pull/90#issuecomment-1783204717).

PR #92, opened 2023-10-31
Add `createIFrameElement` helper

kenzieschmoll:
In trying to migrate from `dart:html` to `package:web`, it was not obvious to me how to create an iFrame element. Not sure if you think this is special enough to special case with its own helper, but thought I'd push up a PR to propose.

@srujzs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 0.3.1-dev
-
+- Add `createIFrameElement` method to `helpers.dart`.
 - Updated types to account for union types.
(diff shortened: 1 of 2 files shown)


kevmoo (changes requested):
Please add a changelog entry!

srujzs (approved):
I'm okay with adding helpers for these as needed, since it's kind of a downgrade from `dart:html` to create elements with string values.

Eventually, we might want to have a helper for each element type and put them in a separate file, but for now, LGTM!

kevmoo (approved):

PR #96, opened 2023-11-10
Prepare for 0.3.1 release.

kenzieschmoll:
Publishing 0.3.1 will unblock https://github.com/flutter/devtools/pull/6626. 
@srujzs @kevmoo


kenzieschmoll:
Even if this can be released as a dev release that is fine - but we need the latest version published to remove the final dart:html dependency from devtools_app

srujzs:
Publishing as a dev release is fine if you want me to go ahead and do that (I think for that you don't even need to land anything, I just need to do a `dart pub publish`).

srujzs (approved):

PR #103, opened 2023-11-22
changelog: drop dev release

kevmoo:
It's just confusing for ~all users

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,9 @@
 ## 0.4.0
 
 - Remove `implements JSObject` from all types. This is needed to move the JS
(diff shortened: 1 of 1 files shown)


srujzs:
While it is confusing, we've already published a `0.3.1-dev` on pub, so I'd keep the section.

kevmoo:
> While it is confusing, we've already published a `0.3.1-dev` on pub, so I'd keep the section.

Yep. But it'll ALWAYS be here: https://pub.dev/packages/web/versions/0.3.1-dev/changelog

For 99.999% of folks, it's just confusing to see the dev release

srujzs (approved):

Issue #105, opened 2023-11-22
Consider excluding IDLs from Web Incubator CG


kevmoo:
We can index off a IDL name into

https://github.com/w3c/webref/blob/main/ed/dfns/

And then look for the spec location.
Might remove a LOT of noise and churn from the generated IDLs!

See 
https://github.com/w3c/webref/blob/d69ef53f2a4ca9b3a58c23bb1d2544f3d12e48cd/ed/dfns/turtledove.json#L4

kevmoo:
Ooo! We can use https://www.npmjs.com/package/web-specs – this has the corresponding names!

CC @srujzs

devoncarew:
Another source of info we could use is the `experimental` flag from the MDN web docs:

https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Experimental_deprecated_obsolete

From that page, features graduate out of experimental if `"[they are] supported by default in two or more major browser rendering engines"`.

devoncarew:
Perhaps close this issue (and track exposing specific APIs in other issues?)?

kevmoo:
> Perhaps close this issue (and track exposing specific APIs in other issues?)?

YES!

srujzs:
I don't want to resolve this one quite yet - I still want to tackle opening up the API surface a bit more (`standard && !experimental`) and the changes needed for that.

PR #111, opened 2023-11-30
parse spec info for generated libraries

devoncarew:
- no longer write generated libraries that wouldn't contain any code
- make the process to clean up previously generated files more lightweight (instead of deleting all the source at the start of generation, we only remove files after generation that hadn't been modified)
- vend in the node `web-specs` package; parse that as part of generation in order to read spec info related to the generated libraries (we don't use this info as yet, but could use it to prune which specs we choose to generate for - see https://github.com/dart-lang/web/issues/105)

This PR is sort of a grab-bag; thoughts welcome.

---

<details>

</details>


kevmoo (approved):
LOVE IT! should wait for 👍  from @srujzs , though

tool/bindings_generator/filesystem_api.dart
```diff
@@ -33,5 +43,19 @@ extension FileSystemExtension on FileSystem {
     }
   }
 
+  @JS('readFileSync')
```
srujzs, line 46: I believe you don't need both, you can make the second argument optional.

Calls to external JS interop functions only pass in the arguments in the call-site, so you still need the two different calls to `readFileSync` if you don't want to pass `null` for `options`, but you don't need both external members.

srujzs (approved):
Nice, thanks! While ideally we should probably move the addition of the web spec to a PR that uses it, I'm totally okay with keeping the processing of the web spec here.

tool/bindings_generator/filesystem_api.dart
devoncarew: I switch this to just:

```
external JSAny readFileSync(JSString path, [JSReadFileOptions options]);
```

and it seems to work fine. Thanks! That's much simpler.

devoncarew:
> Nice, thanks! While ideally we should probably move the addition of the web spec to a PR that uses it,

Yeah, totally - this would normally be the 1st half of a PR. After looking at the info we were parsing however, it didn't immediately jump out at me which libraries not to include in the generation. I think this will need someone w/ more domain expertise to incrementally prune the generated libraries.

We may even want to use the info to generate some libraries into a `src/experimental` directory. It could be nice to make some APIs available, but w/o strong guarantees wrt api stability.

srujzs (approved):

srujzs:
> We may even want to use the info to generate some libraries into a src/experimental directory. It could be nice to make some APIs available, but w/o strong guarantees wrt api stability.

This would be great! One of the concerns I had was that due to the noise of some of the in-progress API definitions, we'd be releasing breaking changes too often. Moving them to a folder where we make no guarantees of stability should hopefully minimize those breaking changes.

kevmoo:
Turns out that info is NOT as useful as I'd hoped. It's too complex to explain here...

PR #116, opened 2023-12-05
Handle Dart SDK change to extension types

srujzs:
This is to make sure code generation doesn't break when we move JS types to extension types. Moving all of the types defined in web.dart to extension types will be handled in a future change.

Updates analyzer dep to 6.3.0 to get the latest extension type changes.


tool/update_bindings.dart
```diff
@@ -175,28 +176,35 @@ Future<void> _generateJsTypeSupertypes() async {
...
   for (final name in definedNames.keys) {
     final element = definedNames[name];
     if (element is TypeDefiningElement) {
+      // TODO(srujzs): This contains code that handles the SDK before and after
+      // the migration of JS types to extension types. Once the changes to
+      // migrate to extension types hit the dev branch, we should remove some of
+      // the old code.
       void storeSupertypes(InterfaceElement element) {
-        bool isInJsTypes(InterfaceElement element) =>
+        bool isInJsTypesOrJsInterop(InterfaceElement element) =>
```
kevmoo, line 190: double-nested function? Necessary? 

Could be moved to top-level?

srujzs: Not necessary, done! I like local functions because they're declared right next to where they're used, but I don't know if that's against dart style.

srujzs:
Friendly ping now that https://dart-review.googlesource.com/c/sdk/+/332860 has merged. :)

PR #121, opened 2023-12-07
fixes to the return types of several Node helper extension methods

devoncarew:
- fixes to the return types of several Node helper extension methods

When trying to port DartPad to package:web, I noticed that the `Node.clone()` extension method had a return type of `void`, which made it hard to use. This PR updates `clone` and `append` to have the same signatures as they have in `dart:html`.

As an aside, when using code completion for a Node element, I see completions for both `append` and `appendChild`, and `clone` and `cloneNode`. Two are from extension methods and two are generated. It's not clear to me as a user which methods I should be using, what the differences are, or why there are multiple options. Perhaps we should plan to deprecate the extension methods which are just there to ease porting from dart:html? Where there are trivially the same generated methods available, and the extension methods don't add any usability improvements?

---

<details>

</details>

### DIFF
--- a/lib/src/helpers/extensions.dart
+++ b/lib/src/helpers/extensions.dart
@@ -7,6 +7,7 @@
 ///
 /// The extensions here are added by hand over time, depending on needs and use
 /// cases. They currently consist of:
+///
 ///  * renames: methods that provide the same functionality, but use a more
 ///    idiomatic Dart name. Typically these renames match the names used in
 ///    `dart:html` in the past.
@@ -66,8 +67,8 @@ extension CanvasRenderingContext2DGlue on CanvasRenderingContext2D {
 
 extension NodeGlue on Node {
   set text(String s) => textContent = s;
-  void append(Node other) => appendChild(other);
-  void clone(bool deep) => cloneNode(deep);
+  Node append(Node other) => appendChild(other);
+  Node clone(bool? deep) => deep == null ? cloneNode() : cloneNode(deep);
 }
 
 extension EventGlue on MouseEvent {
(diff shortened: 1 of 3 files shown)


lib/src/helpers/extensions.dart
```diff
@@ -66,8 +67,8 @@ extension CanvasRenderingContext2DGlue on CanvasRenderingContext2D {
 
 extension NodeGlue on Node {
   set text(String s) => textContent = s;
-  void append(Node other) => appendChild(other);
-  void clone(bool deep) => cloneNode(deep);
+  Node append(Node other) => appendChild(other);
+  Node clone(bool? deep) => deep == null ? cloneNode() : cloneNode(deep);
```
srujzs, line 71: I believe we could rewrite this as `cloneNode(deep ?? false)`. The default value is [`false`](https://dom.spec.whatwg.org/#ref-for-dom-node-clonenode%E2%91%A0).

srujzs:
> Perhaps we should plan to deprecate the extension methods which are just there to ease porting from dart:html? Where there are trivially the same generated methods available, and the extension methods don't add any usability improvements?

cc @sigmundch 

I agree! I think while this might make migration slightly easier, it'll complicate how users use these types. These members are simple enough as well that we can remove them. If you want to remove them in this PR, that's fine by me.

srujzs (approved):

lib/src/helpers/extensions.dart
devoncarew: Updated below:

```suggestion
  Node clone(bool? deep) => cloneNode(deep ?? false);
```

We may also be able to do `... => cloneNode(deep);` ?

srujzs: I actually don't know what passing `null` does! I assume it's considered "falsy" but there's no documentation there.

srujzs:
Sounds good, thanks Devon!

Issue #122, opened 2023-12-07
Add documentation for types and their members


srujzs:
We should likely fetch data from the MDN and auto-annotate IDL types.

devoncarew:
The data from the `web-features` npm package looks like a better source of feature info than the `web-specs` one I started pulling from (https://github.com/dart-lang/web/pull/111).

devoncarew:
This is mostly done; we can track remaining work / improvements in more specific issues (eg. https://github.com/dart-lang/web/issues/190).

PR #130, opened 2023-12-29
collect MDN API information

devoncarew:
- contribute a script which scrapes `https://developer.mozilla.org/en-US/docs/Web` for interface and interface member api documentation
- write that info into a json file in this repo 

This PR collects the info that we'd later want to write into the API dart docs for package:web; some work towards https://github.com/dart-lang/web/issues/122.

---

<details>

</details>


kevmoo:
Oooo!

devoncarew:
It looks like the json file's too large to display as part of the PR. Here's the first few lines of it:

```json
{
  "__meta__": {
    "source": "[MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web)",
    "license": "[CC-BY-SA 2.5](https://creativecommons.org/licenses/by-sa/2.5/)"
  },
  "ANGLE_instanced_arrays": {
    "docs": "The **`ANGLE_instanced_arrays`** extension is part of the WebGL API and allows to draw the same object, or groups of similar objects multiple times, if they share the same vertex data, primitive count and type.\n\nWebGL extensions are available using the `WebGLRenderingContext.getExtension()` method. For more information, see also Using Extensions in the WebGL tutorial.\n\n**Note:** This extension is only available to WebGL1 contexts. In WebGL2, the functionality of this extension is available on the WebGL2 context by default and the constants and methods are available without the \"`ANGLE`\" suffix.\nDespite the name \"ANGLE\", this extension works on any device if the hardware supports it and not just on Windows when using the ANGLE library. \"ANGLE\" just indicates that this extension has been written by the ANGLE library authors.\n\nSee also https://developer.mozilla.org/en-US/docs/Web/API/ANGLE_instanced_arrays."
  },
  "AesCbcParams": {
    "docs": "The **`AesCbcParams`** dictionary of the Web Crypto API represents the object that should be passed as the `algorithm` parameter into `SubtleCrypto.encrypt()`, `SubtleCrypto.decrypt()`, `SubtleCrypto.wrapKey()`, or `SubtleCrypto.unwrapKey()`, when using the AES-CBC algorithm.\n\nSee also https://developer.mozilla.org/en-US/docs/Web/API/AesCbcParams."
  },
  "Accelerometer": {
    "docs": "**Experimental:** **This is an experimental technology**\n\nCheck the Browser compatibility table carefully before using this in production.\n\nThe **`Accelerometer`** interface of the Sensor APIs provides on each reading the acceleration applied to the device along all three axes.\n\nTo use this sensor, the user must grant permission to the `'accelerometer'`, device sensor through the Permissions API.\n\nThis feature may be blocked by a Permissions Policy set on your server.\n\nSee also https://developer.mozilla.org/en-US/docs/Web/API/Accelerometer.",
    "properties": {
      "x": "Returns a double containing the acceleration of the device along the device's x axis.",
      "y": "Returns a double containing the acceleration of the device along the device's y axis.",
      "z": "Returns a double containing the acceleration of the device along the device's z axis."
    }
  },
  "AbortController": {
    "docs": "The **`AbortController`** interface represents a controller object that allows you to abort one or more Web requests as and when desired.\n\nYou can create a new `AbortController` object using the `AbortController()` constructor. Communicating with a DOM request is done using an `AbortSignal` object.\n\nSee also https://developer.mozilla.org/en-US/docs/Web/API/AbortController.",
    "properties": {
      "abort": "Aborts a DOM request before it has completed. This is able to abort fetch requests, consumption of any response bodies, and streams.",
      "signal": "Returns an `AbortSignal` object instance, which can be used to communicate with, or to abort, a DOM request."
    }
  },
...
```

tool/scrape_mdn.dart
```diff
@@ -0,0 +1,239 @@
...
+  }).toList();
+
+  client.close();
+
+  const encoder = JsonEncoder.withIndent('  ');
+
+  final file = File('tool/mdn.json');
+  final json = {
+    '__meta__': {
+      'source': '[MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web)',
```
srujzs, line 54: nit: Can we use the const url from above?

tool/scrape_mdn.dart
```diff
@@ -0,0 +1,239 @@
...
+    for (var i in interfaces) i.name: i.asJson,
+  };
+  file.writeAsStringSync('${encoder.convert(json)}\n');
+}
+
+Future<InterfaceInfo> populateInterfaceInfo(
+  String interfaceName, {
+  required http.Client client,
+}) async {
+  print('  $interfaceName');
```
srujzs, line 67: nit: It feels like printing every interface name will likely get too noisy.

tool/scrape_mdn.dart
```diff
@@ -0,0 +1,239 @@
...
+  final interfaces = await pool.forEach(interfaceNames, (item) async {
+    return populateInterfaceInfo(item, client: client);
+  }).toList();
+
+  client.close();
+
+  const encoder = JsonEncoder.withIndent('  ');
+
+  final file = File('tool/mdn.json');
+  final json = {
```
srujzs, line 52: Is there any benefit to writing out the json and then consuming it later versus just having the node information in memory? If this is to just visualize the contents for now, then ignore this comment.

srujzs (approved):
This is great, thanks Devon!

devoncarew:
It looks like the system-of-record for this information are markdown files (e.g. https://github.com/mdn/content/blob/main/files/en-us/web/api/htmlmediaelement/src/index.md?plain=1). We may want to look into reading from them directly instead of from the served html.

tool/scrape_mdn.dart
devoncarew: I'll switch to just printing something once we're done gathering all the info. The whole script takes around 20 seconds to run.

kevmoo:
> It looks like the system-of-record for this information are markdown files (e.g. https://github.com/mdn/content/blob/main/files/en-us/web/api/htmlmediaelement/src/index.md?plain=1). We may want to look into reading from them directly instead of from the served html.

Big fan of separating download from generate. Makes iterative hacking on the generator a LOT easier.

tool/scrape_mdn.dart
```diff
@@ -0,0 +1,239 @@
...
+    interfaceNames.add(interfaceName);
+  }
+
+  interfaceNames.sort();
+
+  print('${interfaceNames.length} items read from $apiUrl.');
+
+  final pool = Pool(6);
+
+  final interfaces = await pool.forEach(interfaceNames, (item) async {
```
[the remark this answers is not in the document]
devoncarew: No; this change lgtm.

tool/scrape_mdn.dart
srujzs: Ah, I see, I didn't realize it was that slow. Then yes, let's avoid this slowing down iteration of the generation script if it's that significant.

I'm a little concerned on how to keep this updated (both in terms of the the script you added and the docs in the output files) if we aren't running the script to run the docs when we generate code regularly. This may not be a big issue after the original generation, but we're likely to have stale documentation. For this CL and the following ones, it's okay to not worry about this for now.

Issue #139, opened 2024-01-12
investigate `dart fix` renames for this package


devoncarew:
We should investigate using `dart fix` to rename from older dart:html class names to the newer package:web ones. At a first pass, here are the renames (this does not include symbols from dart:svg, ...).

This table has all the changed symbols on the left - or symbols with no match in package:web - and all the matching symbols on the right. It doesn't include the ~300-400 that didn't change.

| `dart:html`                     | `package:web`                   |
| ---                             | ---                             |
| `AbortPaymentEvent`             |                                 |
| `AccessibleNode`                |                                 |
| `AccessibleNodeList`            |                                 |
| `AnchorElement`                 | `HTMLAnchorElement`             |
| `AnimationEffectReadOnly`       |                                 |
| `AnimationEffectTiming`         |                                 |
| `AnimationEffectTimingReadOnly` |                                 |
| `ApplicationCache`              |                                 |
| `ApplicationCacheErrorEvent`    |                                 |
| `AreaElement`                   | `HTMLAreaElement`               |
| `AudioElement`                  | `HTMLAudioElement`              |
| `BackgroundFetchClickEvent`     |                                 |
| `BackgroundFetchedEvent`        |                                 |
| `BackgroundFetchFailEvent`      |                                 |
| `BackgroundFetchFetch`          |                                 |
| `BackgroundFetchSettledFetch`   |                                 |
| `BaseElement`                   | `HTMLBaseElement`               |
| `BluetoothRemoteGattDescriptor` | `BluetoothRemoteGATTDescriptor` |
| `BodyElement`                   | `HTMLBodyElement`               |
| `BRElement`                     | `HTMLBRElement`                 |
| `BudgetState`                   |                                 |
| `ButtonElement`                 | `HTMLButtonElement`             |
| `CanvasElement`                 | `HTMLCanvasElement`             |
| `CDataSection`                  | `CDATASection`                  |
| `ContentElement`                |                                 |
| `Coordinates`                   |                                 |
| `Css`                           | `CSS`                           |
| `CssCharsetRule`                |                                 |
| `CssConditionRule`              | `CSSConditionRule`              |
| `CssFontFaceRule`               | `CSSFontFaceRule`               |
| `CssGroupingRule`               | `CSSGroupingRule`               |
| `CssImageValue`                 | `CSSImageValue`                 |
| `CssImportRule`                 | `CSSImportRule`                 |
| `CssKeyframeRule`               | `CSSKeyframeRule`               |
[...]

PR #141, opened 2024-01-12
contribute a fix_data.yaml file to enable 'dart fix' renames

devoncarew:
- contribute a fix_data.yaml file to enable 'dart fix' renames
- document the class renames from `dart:html` to `package:web`
- related to https://github.com/dart-lang/web/issues/139

@keertip - can you review the fix_data.yaml file? I have some questions about it (in-line).
cc @bwilkerson and @jacob314 as we discussed this yesterday

---

<details>

</details>


example/example.dart
```diff
@@ -5,6 +5,11 @@
 import 'package:web/web.dart';
 
 void main() {
-  final div = document.querySelector('div') as HTMLDivElement;
+  // ignore: unused_local_variable
+  final HtmlDocument foo = document;
```
devoncarew, line 9: This location _does_ offer a quick fix, to rename to `Document`.

example/example.dart
```diff
@@ -5,6 +5,11 @@
...
+
+  final doc = document as HtmlDocument;
```
devoncarew, line 11: This location does _not_ offer a quick fix. Should I define the fix data differently? Is the rename fix expected to cover this case?

cc @keertip

devoncarew:
Somewhat related - and I can open a separate issue for this - quick fix will currently un-do the work of the `dart fix` fixes. When I open the quick fixes available at a ctor reference to an older symbol from dart:html (w/ just package:web in the imports, not dart:html):

Screenshot 2024-01-12 at 8 44 54 AM

The 3rd item above is what we want the user to select - to rename the older symbol name to the new (`HTMLDivElement`) name. However the first item suggested - adding an import for dart:html - will actually work against the migration, and will likely be confusing to casual users. We should probably stop suggesting adding imports for the legacy web libraries (eg, https://api.dart.dev/main/a5758a0dc373e137ee2cf486b5a537fbd5696a95/index.html).

devoncarew:
It would be great to be able to bump the 'rename' quick fix to be first in the list in any case - for it to be suggested ahead of the 'create new class' fix.

parlough:
> and will likely be confusing to casual users. We should probably stop suggesting adding imports for the legacy web libraries

+1000 This is also what I think would be most useful in the migration. Something like https://github.com/dart-lang/sdk/issues/54347

example/example.dart
bwilkerson: I would have expected the rename change to cover this case. It looks like a bug to me.

lib/fix_data.yaml
```diff
@@ -0,0 +1,1342 @@
...
+# format, see https://dart.dev/go/data-driven-fixes.
+
+version: 1
+
+transforms:
+  # AnchorElement => HTMLAnchorElement
+  - title: "Rename to 'HTMLAnchorElement'"
+    date: 2024-01-12
+    element:
+      uris: [ 'package:web/web.dart' ]
```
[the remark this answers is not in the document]
bwilkerson: Either ought to be fine. The shorter form is an "abbreviated URI". See https://github.com/flutter/flutter/wiki/Data-driven-Fixes#uri.

bwilkerson (commented):
Consider defining tests for each of the transformations as per https://github.com/flutter/flutter/wiki/Data-driven-Fixes#testing. I strongly recommend having at least one `.dart` file per transform, even though the flutter team has chosen to put the tests for multiple transforms in a single file. I think it makes it much easier to debug problems when the test files are smaller.

devoncarew:
> Consider defining tests for each of the transformations as per https://github.com/flutter/flutter/wiki/Data-driven-Fixes#testing.

Thanks for the feedback and the idea! I added content in a `test_fixes/` dir. It did find 1/2 dozen quick fixes that were not taking effect - places where we had introduced typedefs to help w/ migrations.

bwilkerson:
Glad we found the bugs.

srujzs (approved):
Awesome, thanks for working on this! I commented on the other issue on why we might be missing some of the mappings. (I'm approving from my perspective - better to have some of the analyzer folks approve before submitting)

lib/fix_data.yaml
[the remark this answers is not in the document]
keertip: Either should be fine. Though we do have this open issue - https://github.com/dart-lang/sdk/issues/52233.

example/example.dart
keertip: As Brian said, it should, otherwise it is a bug. Could you open an issue to capture this?

keertip (approved):
LGTM for the fix_data.yaml file. Please do test that dart fix works if you change the uri from 'package:'.

lib/fix_data.yaml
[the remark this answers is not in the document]
devoncarew: Hmm - is that issue a blocker? Does it mean we should include both types of urls in the fix data file (`web.dart` and `package:web/web.dart`)? Is this is going to affect us it would be great to prioritize that fix for the next release.

devoncarew:
> LGTM for the fix_data.yaml file. Please do test that dart fix works if you change the uri from 'package:'.

I did try updating the fix_data references to use `web.dart` (from `package:web/web.dart`). The fixes didn't work after that - either from the editor (in the example file) or via the cli testing (`--compare-to-golden`) technique. Keeping the longer form for now.

example/example.dart
devoncarew: https://github.com/dart-lang/sdk/issues/54617

PR #146, opened 2024-01-22
Generate interop extension types instead of @staticInterop classes

srujzs:
Also:

- Refactors code around generating the JS type hierarchy
- Updates constraints on code_builder and dart_style
- Refactors code around generating `@JS` annotations


analysis_options.yaml
```diff
@@ -2,6 +2,8 @@
 include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
+  enable-experiment:
+    - inline-class
```
parlough, line 6: I'd likely just update the SDK constraint to `^3.3.0-214.0.dev` or similar instead, which should not require this and is the first public `-dev` release with https://github.com/dart-lang/sdk/commit/b1a7ca77e0272071f41d9eeaadadc52906a1dbc9.

srujzs: I updated the CI to only test `beta` and `dev`, which I believe should both have that version. I'm not sure if we want to update the SDK constraint or not, but I'm leaning towards yes. The benefit of not doing so is users on 3.2 can use this new version, but maybe that's not something we want anyways.

sigmundch: Since we added the package into an allow list to be able to use it early, we can use 3.3.0-76.0.dev (see https://github.com/dart-lang/sdk/commit/f900eb95b784968d04a34bc5f8eeb83afc713aa9), still not a 3.2 unfortuantely.

srujzs:
FYI - it looks like moving the extension members into extension types comes across one issue (`SVGElement.className` is a getter that returns a different type from its setter `Element.className`). I'm still addressing that issue so I'd hold off on a review until the bots are green for this PR.

tool/generator/translator.dart
```diff
@@ -344,7 +345,17 @@ class _PartialInterfacelike {
           if (attribute.special == 'static') {
             staticMembers.add(member);
           } else {
-            members.add(member);
+            if (name == 'SVGElement' && attribute.name == 'className') {
```
kevmoo, line 348: Part of me wants to scream: "generalize this!" but maybe we wait for rule of three.

If we have a couple more special cases, we should probably come up with some `provider` or something to encapsulate these things...

sigmundch (approved):
It's happening! :tada: :partying_face:

.github/workflows/build.yml
```diff
@@ -21,7 +21,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [stable]
+        # TODO(srujzs): Replace with stable when 3.3 is out.
+        sdk: [beta]
```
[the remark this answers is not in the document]
srujzs: Can you elaborate what you mean? Wouldn't any branch testing stable come across analysis issues because the experiment flag is not enabled yet?

tool/generator/translator.dart
sigmundch: Interestingly, in `dart:svg` we hid this getter (we renamed it to `_svgClassName` and didn't expose it elsewhere). We did so with side-car info to the generator, which I guess is not that different than this.

.github/workflows/build.yml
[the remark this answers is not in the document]
sigmundch: I was thinking two branches at different states. Similar to our SDK main/dev/beta/stable, we could have web-main and web-published. The web-published branch would test the code at the current published version of the package. Then we could detect if an SDK changes breaks it.

Today (and this is true for most packages, I believe), we only test the unpublished code at `main`, we may not be aware of issues concerning current users of the package.

I sense this is the wrong CI to test that, though. Maybe a better approach would be to create an "examples" package that uses the stable SDK and regular `dart pub get` to get the current version `package:web` and runs some tests on top of it.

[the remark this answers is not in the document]
srujzs: IIUC, this is kind of another dimension to the matrix where we're testing not only multiple SDKs but multiple package versions? I like it - it'll let us know if we need to mark a previous package as broken, but I'm not sure if the tooling here is setup for that.

> Maybe a better approach would be to create an "examples" package that uses the stable SDK and regular dart pub get to get the current version package:web and runs some tests on top of it.

That would work as a workaround. It'd be nice to have a package test multiple versions of a dependency in this file as well. This would have been useful when we were dealing with the `>=0.3.0 <0.5.0` constraints Flutter packages had to use.

tool/generator/translator.dart
srujzs: That should show you how long it's been deprecated but not removed. :)

I considered not exposing it, but that made less sense because that is indeed what the DOM exposes. If we decide to avoid generating deprecating members in this package as part of our versioning decisions, this special-casing can be removed.

PR #150, opened 2024-01-22
API diffs for several IDL versions (from `3.37.0` to `3.42.0`)

devoncarew:


lib/src/dom/html.dart
```diff
@@ -4166,18 +4174,24 @@ class PopStateEvent implements Event {
...
 }
 
 @JS()
 @staticInterop
 @anonymous
 class PopStateEventInit implements EventInit {
-  external factory PopStateEventInit({JSAny? state});
+  external factory PopStateEventInit({
+    JSAny? state,
+    bool hasUAVisualTransition,
```
devoncarew, line 4231: @srujzs - should we be generating this as `bool? hasUAVisualTransition`? `required bool hasUAVisualTransition`? I assume the `bool` param is effectively optional.

srujzs: https://github.com/w3c/webref/blob/b8cdf28506d6699133866dbe950ff20873d84218/ed/idl/html.idl#L2045

`external` members are weird. You're allowed to write non-nullable members with no default to communicate that JS has its own default. This is probably better than `bool?` because it disallows users from passing a `null` value.

devoncarew:
closing as this was just for comparison purposes

PR #153, opened 2024-01-31
Add generics to JSArrays and JSPromises

srujzs:
Adds a "typeParameter" field to _RawType that keeps track of the generic type for union and emission purposes. Also handles typedefs so that the generic type is always a valid interop type and not a Dart primitive type.

The typedef handling comes at the cost of not having a better name for these types when they appear in generics and some typedefs being potentially unused. The alternative is to make all typedefs JS types but that means an increased need for conversions. From a quick analysis, this would affect several hundred APIs. Another alternative is to have two typedefs, one for the generic and one for the non-generic. This has its own visual downsides as we now have two typedefs for every typedef.


sigmundch (approved):
Looks great!

PR #156, opened 2024-02-01
Generate library declarations with @JS annotations

srujzs:
This is needed to work around https://github.com/dart-lang/sdk/issues/54801. When that issue is fixed, these declarations can be removed.


lib/src/dom/anchors.dart
```diff
@@ -4,6 +4,9 @@
 
 // Generated from Web IDL definitions.
 
+@JS()
+library anchors;
```
kevmoo, line 8: no name needed! just `library;` is great!

kevmoo (changes requested):

lib/src/dom/anchors.dart
srujzs: Done!

lib/src/dom/xhr.dart
```diff
@@ -4,6 +4,9 @@
 
 // Generated from Web IDL definitions.
 
+@JS()
+library;
```
sigmundch, line 8: curious, is the `library;` piece needed? I have this recollection that the first annotation before any import is still associated with the library, but I may be wrong.

sigmundch (approved):
Thanks, feel free to ignore my question, it's probably not worth while anyways.

lib/src/dom/xhr.dart
srujzs: You might be right (at least it compiles and works in this case), TIL. However, `code_builder` explicitly adds the `library`: https://github.com/dart-lang/code_builder/blob/3307105e57a1cbcc08faff250800bf3b93fbfe15/lib/src/emitter.dart#L521. I'm guessing they also didn't know this was possible!

kevmoo (approved):

lib/src/dom/xhr.dart
parlough: I think that makes a lot of sense!

To ease the transition we could add https://dart.dev/tools/linter-rules/library_annotations to the core lints.

eernstg: Thanks! And thanks for the supporting arguments! ;-)

sigmundch: I can be convinced of such change :). At a high level, I'd lean towards what's more common. If it's more common to use library annotations than import annotations, then it's nice not to have the `library;` directive. I agree, though, that when you do need to put an annotation on the first import, it will feed strange that you need to add a `library` directive above it to make it work.

sigmundch: The thing that helps here is that dart-doc comments have the same behavior. That tends to make the annotation quicks less surprising.

kevmoo: Yes! We had a bunch of vagueness about this. it's a bit more verbose, but it's 100% clear to have the library directive to hang things on!

PR #157, opened 2024-02-01
Change records to be JSObject

srujzs:
Records are key-value pairs that are represented as objects in JavaScript. https://webidl.spec.whatwg.org/#idl-record

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   primitive type are instead replaced with their JS type equivalent if they
   appear as a type parameter in order to conform with the type bounds of
   `JSArray` and `JSPromise`.
+- Changed `record` types to be `JSObject` instead of `JSAny`.
 
 ## 0.4.2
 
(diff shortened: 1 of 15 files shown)


kevmoo:
It'd be nice to expose a `Map`-looking thing for these...at some point

srujzs:
We could borrow from TypeScript and have our own Record JS type: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type :)

Its Dart equivalent could just be `Map` (although Dart's `Map` could also apply to ES6 Maps), but like with any other extension type, there is no strong guarantee that the JS type's type parameters are accurate.

sigmundch (approved):
Is there any info about the fields on these records? If so, would it make sense to declare an object-literal interop class (rather than a  Map)?

srujzs:
> Is there any info about the fields on these records? If so, would it make sense to declare an object-literal interop class (rather than a Map)?

No, records are represented only by the generic types associated with their key and value e.g. `record<DOMString, DOMString>`, whereas dictionaries specify the name of the field. The keys can likely be a very wide range of values e.g. https://w3c.github.io/clipboard-apis/#dom-clipboarditem-clipboarditem.

PR #162, opened 2024-02-07
Remove helper typedefs in favor of fixes

parlough:
Now that developers don't have a choice to opt in to the `helpers.dart` members being available, I think these type definitions are detrimental to the developer experience. They increase the items in the global scope, introduce more choices, and increase inconsistency.

Let's help developers of existing packages/apps get to a consistent end-state sooner by leaning on data-driven fixes and help developers working on new packages/apps by having a more streamlined experience.


devoncarew:
I'll let @srujzs (and @kevmoo and @sigmundch) speak to the desired API here, but do agree that the data-driven fixes give us more options in terms of migrating from dart:html => package:web. We can lean on data-driven fixes - which will patch up code with deprecation messages or missing type references - instead of having a transitory shim layer. Perhaps there's a use case for both? But it may be a clearer story to change the import statement and fix-up all code to the new package:web API.

srujzs:
Now that we have the dart fix, I don't see much value in this anymore. It'll lead to confusing conflicts, and it's better to deprecate them now before people start relying on them.

srujzs (approved):

PR #163, opened 2024-02-07
Deprecate `HttpRequest` helper

parlough:
Deprecate this helper to prevent significant future use of it as it's no longer gated behind a `helpers.dart` library.

Just like the `HttpRequest` before it, we should direct users to `package:http` instead for a consistent, cross-platform developer experience.

If developers don't want to bring in a dependency in `package:http`, they can still use `XMLHttpRequest` or `fetch` directly.

\cc @brianquinlan


kevmoo:
I think I agree here. Thoughts @devoncarew ?

brianquinlan:
I support more promotion of `package:http` ;-)

devoncarew:
> I think I agree here. Thoughts @devoncarew ?

It looks like it's just a convenience wrapper on top of the package:web API? I suspect they should use package:web `XMLHttpRequest` directly - or `fetch` - or the x-platform package:http wrapper. It would be good to avoid another wrapper / abstraction here.

srujzs (approved):

srujzs:
LGTM, going to land it ahead of 0.5.0.

PR #164, opened 2024-02-07
Deprecate extension methods returning `Point`

parlough:
`Point` really shouldn't exist in its current form so we shouldn't encourage or enable its use.

If anyone still thinks these extensions are helpful, maybe we could return a record instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 - Changed `record` types to be `JSObject` instead of `JSAny`.
 - Reduce the number of DOM APIs we generate code for. Currently, the API needs
   to be standards-track, and be suported by Safari, Chrome, and Firefox.
+- Deprecate the `client` extension methods on `MouseEvent` and `Touch`.
+  Instead, directly use the `clientX` and `clientY` properties.
 
 ## 0.4.2
 
--- a/lib/src/helpers/extensions.dart
+++ b/lib/src/helpers/extensions.dart
@@ -74,10 +74,22 @@ extension NodeGlue on Node {
 }
 
 extension EventGlue on MouseEvent {
+  /// A [Point] representation of the [clientX] and [clientY] properties
+  /// of this [MouseEvent].
+  ///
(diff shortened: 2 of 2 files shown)


kevmoo:
Ooo....record could be nice.

kevmoo:
@srujzs @devoncarew – thoughts here?

kevmoo:
Why don't we like `Point`, @parlough ?

srujzs:
I am curious why we want to discourage `Point` as well - is `dart:math` planning on deprecating it?

parlough:
A `Point` on a screen does not really belong in a math library, nor does `Rectangle`. Beyond that, `Point` declares operators that don't make sense for a point, conflating it with a vector. The `*` operator on `Point<int>` can also have a runtime error 😢 when provided with a `double`.

Beyond those potential sources of confusion, I guess it being generic can introduce performance issues (https://github.com/dart-lang/sdk/issues/53912).

I believe there's been a variety of other discussions around this over the years, but since `dart:html` used it, it was likely hard to deprecate and remove. With this web migration, I think it's time to revisit that. Whether we do/can or not, it seems better avoid new uses where possible. Anyone using `Point` here likely just needs the x and y, and direct access or a record can provide that.

srujzs:
Thanks for the reasoning! I'm okay with avoiding new uses for now. We can always re-add (or users can in their own libraries) if we think this is safe.

srujzs (approved):

kevmoo:
Or do we just move to the `Record` and include it in the breaking change?!?!

Might be good to do now!

srujzs:
Either way works. I presume there's very low usage of this, and since we're releasing a breaking change version, might as well remove it instead of deprecating.

kevmoo:
@srujzs – or just replace the return type!

parlough:
I'd really suggest we deprecate or remove these for this release. Then revisit extensions returning records in the future. For example `MouseEvent` has a bunch of other `somethingX` and `somethingY` values that we don't expose through extensions like this.

PR #174, opened 2024-02-15
Emit all APIs that are part of the standards track

srujzs:
(purely testing to see the diff here)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
         # TODO(srujzs): Replace with stable when 3.3 is out.
         sdk: [dev, 3.3]
     steps:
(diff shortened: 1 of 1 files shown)


lib/src/dom/anchors.dart
```diff
@@ -0,0 +1,19 @@
...
+// Generated from Web IDL definitions.
+
+@JS()
+library;
+
+import 'dart:js_interop';
+
+import 'webxr.dart';
+
+extension type XRAnchor._(JSObject _) implements JSObject {
```
kevmoo, line 14: https://caniuse.com/mdn-api_xranchor

Still considered experimental 🤷

srujzs: Yup, another experiment I did locally was `standardsTrack && !experimental`. That was a net of +2000 or so and includes Trusted Types.

Issue #178, opened 2024-02-19
is RTCLocalSessionDescriptionInit needed?


jarrodcolburn:
Problem: `RTCLocalSessionDescriptionInit` class for "local" seems unneeded. I think `RTCSessionDescriptionInit` would suffice. Please see the screenshot where inconsistent between local and remote is highlighted

![image](https://github.com/dart-lang/web/assets/16673615/fe14b154-5e7b-4e79-b10c-dca588f2aef7)

<hr>
Recommendation A. Get rid of `RTCLocalSessionDescriptionInit`

Recommendation B (less good)...  If `RTCLocalSessionDescriptionInit` is needed... it should probably extend `RTCSessionDescriptionInit` and perhaps a remote version `RTCRemoteSessionDescriptionInit` should exist for consistency with local and remote.

jarrodcolburn:
for more context, please see https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createOffer
```javascript
myPeerConnection
  .createOffer()
  .then((offer) => myPeerConnection.setLocalDescription(offer))
```
This Dart equivalent of this js won't work because in Dart `createOffer()` returns a "non-local" description, but Dart's `setLocalDescription` requires "local" description.

jarrodcolburn:
Current implementation and links:
```dart
extension type RTCSessionDescriptionInit._(JSObject _) implements JSObject {
  external factory RTCSessionDescriptionInit({
    required RTCSdpType type,
    String sdp,
  });
```

```dart
extension type RTCLocalSessionDescriptionInit._(JSObject _) implements JSObject {
  external factory RTCLocalSessionDescriptionInit({
    RTCSdpType type,
    String sdp,
  });
```

[RTCSessionDescriptionInit](https://github.com/dart-lang/web/blob/975e55cbe4ac683c33d5256a91f4133fa97b4c34/lib/src/dom/webrtc.dart#L194)


[RTCLocalSessionDescriptionInit](https://github.com/dart-lang/web/blob/975e55cbe4ac683c33d5256a91f4133fa97b4c34/lib/src/dom/webrtc.dart#L205)

jarrodcolburn:
Opened a ticket on the webref repo https://github.com/w3c/webref/issues/1155

kevmoo:
> Opened a ticket on the webref repo [w3c/webref#1155](https://github.com/w3c/webref/issues/1155)

[...]

PR #179, opened 2024-02-20
Fix: deprecate local for rtcsession

jarrodcolburn:
see https://github.com/dart-lang/web/issues/178

### DIFF
--- a/lib/src/dom/webrtc.dart
+++ b/lib/src/dom/webrtc.dart
@@ -117,7 +117,7 @@ extension type RTCPeerConnection._(JSObject _)
     RTCPeerConnectionErrorCallback failureCallback,
   ]);
   external JSPromise<JSAny?> setLocalDescription([
-    RTCLocalSessionDescriptionInit description,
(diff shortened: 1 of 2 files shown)


jarrodcolburn:
In https://github.com/dart-lang/web/issues/178 I suggest two recommendations. This is a PR that implements A, the first, to remove "local". 
This also includes a small test which refers to the MDN link mentioned.

kevmoo:
Love it...BUT! These APIs are auto-generated. We'll need to figure out how to create exceptions or overrides or such

jarrodcolburn:
Ah. It's not breaking my app, but it is causing me to have to do an unnecessary step in my code, hurting readability for anyone coming behind me. Working with RTC already hurts my brain

jarrodcolburn:
Although, you're probably right "exceptions or overrides" will be needed at some point to tweak things. For this, I'm going to try to open a ticket on the actual definitions being used for auto generation, I guess that would be on the https://github.com/w3c/webref repo, right?

PR #184, opened 2024-02-23
Use extension types and generics internally

srujzs:
Migrate @staticInterop to extension types, which enables us to use generic arguments to JSArray and JSPromise.


kevmoo (approved):
LGTM

PR #185, opened 2024-02-23
Add generated element constructors

srujzs:
For every tag that the interface supports, emits a Dart constructor that can be used to create an element of that tag using either createElement or createElementNS. Adds a dependency to @webref/elements to fetch the element data. Modifies existing helper deprecation messages to point to the new constructors instead. Deprecates createAudioElement.

As we remove some members that don't work, and this deprecation is a breaking change, we update the major version number.


srujzs:
FYI @kevmoo - This requires moving to `0.6.X`.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
-## 0.5.1-wip
+## 0.6.0-wip
```
kevmoo, line 1: must it be a breaking change? That's...expensive

srujzs: There's two breaking changes here:

1. Removing the already broken `external` constructors. I believe even though they were broken, making them static errors constitutes a breaking change, but maybe I'm wrong.
2. Deprecating `createAudioElement`. This could potentially break Flutter if we don't roll the version number, although we could certainly defer the deprecation until later.

srujzs: The language is a bit vague here, but we may be able to get away without rolling a breaking change since these members were already broken: https://github.com/dart-lang/sdk/blob/main/docs/process/breaking-changes.md#dart-compatibility-and-philosophy.

> Implementation bugs: If the implementation deviates unintentionally from the specification, programs may break as we rectify the implementation.

kevmoo: deprecations are no longer breaking for flutter. Or we could just put in an ignore.

srujzs: Ooh, can you link me to where that has changed? I've been following this: https://github.com/flutter/flutter/issues/108555 but there hasn't been any updates there. If that's the case, then we can just remove the version bump here, and if not, we can just defer the deprecation until that changes.

kevmoo: @matanlurey ?

We could also just change the code we need to before we roll this.

matanlurey: I don't have anything better to point to, but see https://github.com/flutter/flutter/issues/143312.

srujzs: Thanks Matan! Looks like the lint was added to the ignore list in all the relevant repos and Flutter doesn't use `createAudioElement` anywhere anyways, so I'll remove the version bump.

srujzs:
> I didn't do the create-a-PR-against-a-PR thing, but you might want the last two file changes from https://github.com/srujzs/web/commit/153424a0a44f27cd79f7ceb3e0b32423c53b56e3.

Thanks Devon! Unfortunately, it's a little annoying to determine what article is correct even ignoring proper nouns, because words like `user` and `hour` defy the vowel rule. For now, I just added a method to only handle the expected inputs. A more general solution is going to be harder: https://stackoverflow.com/questions/4558437/programmatically-determine-whether-to-describe-an-object-with-a-or-an.

I also got rid of the wrapping method I added and used the one you added with no changes to the output.

kevmoo (approved):

sigmundch (commented):
Finally finished taking a look! LGTM too :)

PR #189, opened 2024-02-28
Move IDL version details

kevmoo:
etc to tool directory,Offers very little benefit for almost all users

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,3 +66,9 @@ definitions:
   that is standards track and supported on Chrome, Firefox, and Safari to reduce
   the number of breaking changes. This is currently WIP and some members may be
   added or removed.
+
+### Generation
+
+Most of the APIs in this package are generated from public assets.
+See the [tool directory](https://github.com/dart-lang/web/tree/main/tool) in the
+source repository to learn more.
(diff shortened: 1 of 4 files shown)


devoncarew:
It's true this may be providing too much info to users? I do think there's some value in letting them know that the package is generated from the IDL, and perhaps value in letting them know how current it is / how frequently it's updated? (i.e., `@webref/idl` / `3.43.1`).

(@srujzs is probably the best reviewer for this)

srujzs:
I agree with Devon that it's useful to include this information in the main README. Having the direct source of truth immediately visible makes the generation seem less magical and makes it easy to find what has changed when we update the IDL version.

kevmoo:
It's still SUPER easy to see the delta in the diff – local or on GitHub.

But it offers very little benefit to the user. And the link doesn't work on `pub.dev`, etc.

srujzs:
It looks like they work here unless I'm missing some other context: https://pub.dev/packages/web.

If we do want to land this, maybe add a comment in the main readme to point to the readme in `tool/` e.g. `See tool/README.md for more details on what Web IDL version the tool uses and how to generate the bindings.`?

kevmoo:
Not linked correctly here though; https://pub.dev/documentation/web/latest/

maybe that's just a bug

srujzs:
I might be missing something again because those links still work for me. :D

kevmoo:
PTAL

srujzs (approved):
Thanks!

PR #217, opened 2024-03-31
Fix document.documentElement should return HtmlHtmlElement #214

### LINKED ISSUE #214: document.documentElement  should return HtmlHtmlElement
`document.documentElement` returns `Element?` but it should return [`HtmlHtmlElement?`](https://pub.dev/documentation/web/latest/web/HTMLHtmlElement-extension-type.html) according to [mdn's docs on HTMLHtmlElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHtmlElement)

> The HTMLHtmlElement interface serves as the root node for a given HTML document. This object inherits the properties and methods described in the [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) interface.

 > You can retrieve the HTMLHtmlElement object for a given document by reading the value of the [document.documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement) property.

ksw2000:
- Fix #214

### DIFF
--- a/lib/src/dom/dom.dart
+++ b/lib/src/dom/dom.dart
@@ -1809,9 +1809,9 @@ extension type Document._(JSObject _) implements Node, JSObject {
   external DocumentType? get doctype;
 
   /// **`Document.documentElement`** returns the
-  /// [Element] that is the root element of the [document] (for
+  /// [HTMLHtmlElement] that is the root element of the [document] (for
   /// example, the `html` element for HTML documents).
-  external Element? get documentElement;
+  external HTMLHtmlElement? get documentElement;
 
   /// The **`Document.location`** read-only property returns a
   /// [Location] object, which contains information about the URL of the


kevmoo:
These APIs are generated. We don't want to patch things in this way.

CC @srujzs

ksw2000:
I think you're right; this issue seems a bit strange.

jarrodcolburn:
You're saying

> The MDN notes also mention that it can be other types of `Element`s: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement#notes.

at link^
> Document.documentElement returns the [Element]

...which it is an Element. But I'd say a `super` type, not an `other` type.

Especially since https://developer.mozilla.org/en-US/docs/Web/API/HTMLHtmlElement says 
> You can retrieve the HTMLHtmlElement object for a given document by reading the value of the [document.documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement) property.

srujzs:
Sorry, I wasn't clear what I was referring to there. Specifically this sentence:

> For any non-empty XML document, documentElement will always be whatever element is the root element of the document.

This sentence leads me to believe this getter can return other types of `Element`s beyond just html elements, so if we made this `HTMLHtmlElement`, it might be wrong in some cases.

Issue #220, opened 2024-04-06
Event Handler's Getters and Setters in dom.dart are defined as smallcase. 


kaljitism:
In `lib/src/dom/dom.dart` Event Handlers getters and setters APIs are Small Case Identifiers, they should be converted to Camel Case Identifiers. 

Code Link - https://github.com/dart-lang/web/blob/main/lib/src/dom/dom.dart

For example - 

```
. . . . . . 
 external set onanimationcancel(EventHandler value);
 external EventHandler get ontransitionrun;
 external set ontransitionrun(EventHandler value);
 external EventHandler get ontransitionstart;
 external set ontransitionstart(EventHandler value);
 external EventHandler get ontransitionend;
 external set ontransitionend(EventHandler value);
 external EventHandler get ontransitioncancel;
 external set ontransitioncancel(EventHandler value);
. . . . . 
```

should be renamed to 

```
. . . . . . 
 external set onAnimationCancel(EventHandler value);
 external EventHandler get onTransitionRun;
 external set onTransitionRun(EventHandler value);
 external EventHandler get onTransitionStart;
 external set onTransitionStart(EventHandler value);
 external EventHandler get onTransitionEnd;
 external set onTransitionEnd(EventHandler value);
 external EventHandler get onTransitionCancel;
 external set onTransitionCancel(EventHandler value);
. . . . . 
```

kaljitism:
I have forked the repo and working on it.

kaljitism:
Fixed this in [PR](https://github.com/dart-lang/web/pull/221)

ykmnkmi:
Did you test all these changes?

kaljitism:
@ykmnkmi Since its a typo related fix, I believe it might be a breaking change. Correct me if I am wrong. 

Also, I did not test. how do I do it for these specific changes??  Would I need to write my own tests or the checks that will be done by Github Actions and Reviewers would be fine?

ykmnkmi:
@kaljitism it's not a typo; the current naming for event handlers directly mirrors that of the JavaScript DOM API. At least add JS annotations with proper names:
```dart
@JS('onanimationcancel')
external set onAnimationCancel(EventHandler value);
```

kaljitism:
I see. On it.

kevmoo:
This is by design. We map exactly to what's in the Web IDL.

PR #221, opened 2024-04-06
DOM Event handler Typo Fix

kaljitism:
In lib/src/dom/dom.dart Event Handlers getters and setters APIs are Small Case Identifiers, they should be converted to Camel Case Identifiers.

Code Link - https://github.com/dart-lang/web/blob/main/lib/src/dom/dom.dart

Related Issue - https://github.com/dart-lang/web/issues/220

---

I have fixed all the APIs in dom.dart to Camel Case according to Style Guide.


kevmoo:
These files are generated from the IDL.

PR #244, opened 2024-05-22
Update MDN source and doc comments

kevmoo:


lib/src/dom/selection_api.dart
```diff
@@ -106,9 +117,12 @@ extension type Selection._(JSObject _) implements JSObject {
...
   ]);
 
-  /// The **`setBaseAndExtent()`** method of the
-  /// [Selection] interface sets the selection to be a range including all or
-  /// parts of two specified DOM nodes, and any content located between them.
+  /// The **`setBaseAndExtent()`** method of the [Selection] interface sets the
+  /// selection to be a range including all or parts of two specified DOM nodes,
+  /// and any content located between them.
+  ///
+  /// The anchor and focus nodes can be located in a , if supported by the
```
srujzs, line 124: Seems like we might not be processing links correctly: https://developer.mozilla.org/en-US/docs/Web/API/Selection/setBaseAndExtent.

devoncarew: Yes, likely we don't support this particular construct:

```
... can be located in a {{glossary("shadow tree")}}, if supported by ...
```

devoncarew (approved):

lib/src/dom/selection_api.dart
devoncarew: https://github.com/dart-lang/web/issues/245

PR #246, opened 2024-05-28
Expose `onUnload` event stream and `ElementStream` as a public class

kenzieschmoll:


srujzs (approved):

srujzs:
Still LGTM after exposing `ElementStream`!

PR #256, opened 2024-06-17
Emit double as return value instead of num

srujzs:
Closes https://github.com/dart-lang/web/issues/57

In cases where the return type is known to be a double value, prefer double over num so users don't try and downcast the type, which won't work consistently on all compilers. In order to do this, _typeReference is amended to take an extra option 'returnType' which determines if the `_RawType` is used in a return type context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@
   dictionaries and typedefs are only emitted if they're used by a generated API.
 - Added `onUnload` event stream to `ElementEventGetters` extension methods.
 - Expose `ElementStream` as a public class.
+- APIs that return a double value now return `double` instead of `num`. This is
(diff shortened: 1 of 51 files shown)


kevmoo:
Cool!

sigmundch (approved):
Nice!

devoncarew:
FYI that when rolling this into google3 we ran into failures w/ some new diagnostics (unnecessary cast) in the devtools codebase. This has been addressed upstream in devtools, so `package:web` can roll in after the next devtools roll (cc @kenzieschmoll).

kenzieschmoll:
CC @elliette who is on the DevTools release / roll rotation this month.

elliette:
The DevTools roll into g3 was just submitted, so package:web should now be unblocked.

devoncarew:
Thanks!

PR #257, opened 2024-06-17
Account for typedef nullability

srujzs:
_desugarTypedef accepts a _RawType and returns its underlying type if it's a typedef. However, it doesn't account for whether the _RawType is nullable or not. In order to accurately desugar a use of a typedef, we should union the _RawType's nullability with the underlying type's nullability. For example, `SomeTypedef?` should always be nullable after being desugared.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Added `onUnload` event stream to `ElementEventGetters` extension methods.
 - Expose `ElementStream` as a public class.
 - Require Dart `^3.4.0`.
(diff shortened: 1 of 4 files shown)


sigmundch (approved):

sigmundch (approved):
(consider also a CHANGELOG entry to indicate why some APIs are changing. Seems that all changes were in contravariant positions, so the change appears non-breaking?)

srujzs:
It is indeed non-breaking. Done, thanks!

PR #265, opened 2024-06-27
Rename members when they collide with return/parameter type names

srujzs:
Web IDL allows members to have the same name as their return or parameter types. Dart does not, and therefore, we have to rename them. Cleans up some existing code and makes renaming consistent by always adding an underscore if needed.

### DIFF
--- a/tool/generator/translator.dart
+++ b/tool/generator/translator.dart
@@ -499,12 +499,11 @@ class _PartialInterfacelike {
             if (staticOperations.containsKey(operationName)) {
               staticOperations[operationName]!.update(operation);
             } else {
-              final staticOperation = _OverridableOperation(
+              staticOperations[operationName] = _OverridableOperation(
                   operation, _MemberName(operationName), docs);
               if (operations.containsKey(operationName)) {
-                staticOperation.underscoreName();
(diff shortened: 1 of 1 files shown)


kevmoo (approved):
Guessing this is tested with the latest idl bits?

srujzs:
Yup, here's the diff from the update to this CL + the update:

```
--- a/lib/src/dom/saa_non_cookie_storage.dart
+++ b/lib/src/dom/saa_non_cookie_storage.dart
@@ -45,10 +45,12 @@ extension type StorageAccessHandle._(JSObject _) implements JSObject {
   external void revokeObjectURL(String url);

   /// > **Note:** See [BroadcastChannel.BroadcastChannel] to understand usage.
-  external BroadcastChannel BroadcastChannel(String name);
+  @JS('BroadcastChannel')
+  external BroadcastChannel BroadcastChannel_(String name);

   /// > **Note:** See [SharedWorker.SharedWorker] to understand usage.
-  external SharedWorker SharedWorker(
+  @JS('SharedWorker')
+  external SharedWorker SharedWorker_(
     String scriptURL, [
     JSAny options,
   ]);
```

PR #267, opened 2024-06-27
Account for static members in MDN docs

srujzs:
Static MDN members are suffixed with `_static`. Like with compatibility data, we should add this suffix when emitting docs.


kevmoo (approved):
Sweet! good catch!

PR #278, opened 2024-08-12
Ignore unintended_html_in_doc_comment

srujzs:
MDN docs includes some variants of html tags that this rule disallows, like `<table class="properties">`, `<th scope="row">`, and `<br />`. In order to make the CI pass, this is ignored for now. Fixes cases unrelated to the MDN docs.

cc @kallentu for some fun edge cases from the MDN.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -22,6 +22,7 @@ analyzer:
     unreachable_from_main: ignore
     # MDN comments include HTML docs that trigger this lint. Reenable once
     # that's resolved.
+    # ignore: unrecognized_error_code
     unintended_html_in_doc_comment: ignore
 
 linter:
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -20,10 +20,6 @@ analyzer:
     non_constant_identifier_names: ignore
     # Consider removing from dart_flutter_team_lints.
     unreachable_from_main: ignore
-    # MDN comments include HTML docs that trigger this lint. Reenable once
-    # that's resolved.
-    # ignore: unrecognized_error_code
-    unintended_html_in_doc_comment: ignore
 
 linter:
   rules:
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -23,18 +23,21 @@ analyzer:
 
 linter:
   rules:
-  - avoid_bool_literals_in_conditional_expressions
-  - avoid_private_typedef_functions
-  - avoid_redundant_argument_values
-  - avoid_returning_this
-  - avoid_unused_constructor_parameters
-  - cancel_subscriptions
-  - join_return_with_assignment
-  - literal_only_boolean_expressions
-  - no_adjacent_strings_in_list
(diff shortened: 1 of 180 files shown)


analysis_options.yaml
```diff
@@ -23,18 +23,21 @@ analyzer:
...
-  - join_return_with_assignment
-  - literal_only_boolean_expressions
-  - no_adjacent_strings_in_list
-  - no_runtimeType_toString
-  - package_api_docs
-  - prefer_const_declarations
-  - prefer_final_locals
-  - unnecessary_await_in_return
-  - use_string_buffers
+    avoid_bool_literals_in_conditional_expressions: true
```
kevmoo, line 26: Instead, use https://dart.dev/tools/analysis#ignoring-rules

```yaml
analyzer:
  errors:
    unintended_html_in_doc_comment: ignore
```

tis cleaner

kevmoo (changes requested):

analysis_options.yaml
srujzs: Done, thanks!

analysis_options.yaml
```diff
@@ -20,6 +20,9 @@ analyzer:
     non_constant_identifier_names: ignore
     # Consider removing from dart_flutter_team_lints.
     unreachable_from_main: ignore
+    # MDN comments include HTML docs that trigger this lint. Reenable once
+    # that's resolved.
```
kevmoo, line 24: nit: should have a bug to reference here?

kevmoo:
Re the failure: we shouldn't run analyzer tests against 3.4 - just dev

kevmoo:
nit: also: could we just add this ignore at the top of all generated files?

srujzs:
> Re the failure: we shouldn't run analyzer tests against 3.4 - just dev

We should still make sure analysis passes with 3.4 so we don't release a package that breaks against an older Dart SDK version. We might need to ignore `unrecognized_error_code` to make this work however.

> nit: also: could we just add this ignore at the top of all generated files?

Yeah, since it looks like we can't exclude files from a lint, I could try that.

analysis_options.yaml
srujzs: https://github.com/dart-lang/sdk/issues/56450 Done!

srujzs:
Generated in the files themselves instead, which has the nice side-effect of it not affecting the 3.4 analysis.

kevmoo (approved):

devoncarew:
I'm trying to validate @kallentu's fix for this, but am having trouble getting the false positives from `unintended_html_in_doc_comment` to fire here (using an sdk before the fix). I'm assuming I have a local config issue but mentioning here as I don't see the false positives.

devoncarew:
Nevermind - I had to check out an older commit to get this to trigger; I do see these as fixed in head.

srujzs:
Cool, thanks Kallen! We'll still need the ignores for a little while until we move to a minimum SDK version that contains the fix.

PR #282, opened 2024-08-17
[feat] Adds http status codes (Resolves #280)

### LINKED ISSUE #280: Add HttpStatus codes api seen in dart html library
### **Issue:**

Dart `html` library exports `HttpStatus` abstract class that defines various http status codes. I can't find a replacement for this in the `web` package. Currently, the `HttpStatus` is part of darts internal apis so one can't directly import it in their project but this api is exported only through dart `html` and `io` libraries. As `web` is replacing the `html` library eventually, addition of this api would be quite handy. 

### **Proposal:**

Add `HttpStatus` api similar to one exported from `html` library. I believe one can just copy the api from dart sdk which should be good enough. I'm ready to put up a PR on this.

### LINKED ISSUE #281: [feat] Add helper to convert HtmlCollection to a dart list
### **Issue:** 

When working with html element childrens which are represented through the `HtmlCollection`, if someone wants to iterate through these childrens in dart, only approach is to use a  `for` loop like follow to either create a dart list which would then allow running operations like `forEach`, `map`, etc on the list or simply iterate through them to execute any functions. 

```dart
      final childrens = <Element>[];

      for (var i = 0; i < element.children.length; i++) {
        final child = element.children.item(i);

        childrens.add(child);
      }
```

This works but it becomes tedious when I've to do the same thing in multiple places. Off course, I could add an extension for this on `HtmlCollection` but it would be nice if such helper was present in `web` package itself. 

`TouchListConvert` extension on `TouchList` is an example of such helper. And the approach seems better instead of running a for loop which results in extra computations. Adding a similar helper for converting `HtmlCollection` to a dart list would be nice addition I believe. 

### **Proposal:** 

I can see this addition going two ways, 

1. Add a similar extension on `HtmlCollection` like `TouchListConvert` that would return a `HtmlCollectionWrapper` that would be a dart list.
2. Create a generic wrapper like `TouchListConvert` which could work for both `TouchListConvert` and `HtmlCollection`, and also other web facing apis like `NodeList` that represent a list of items. This one I'm not so sure abt how one would approach for this scenario.

rutvik110:
[feat] Adds http status codes (Resolves #280)

---

<details>

</details>

### DIFF
--- a/lib/src/helpers/enums.dart
+++ b/lib/src/helpers/enums.dart
@@ -227,3 +227,71 @@ abstract final class Device {
   static final String propertyPrefix =
       isFirefox ? 'moz' : (isIE ? 'ms' : (isOpera ? 'o' : 'webkit'));
 }
+
+/// HTTP status codes.
+abstract class HttpStatus {
+  static const int continue_ = 100;
+  static const int switchingProtocols = 101;
+  static const int processing = 102;
+  static const int ok = 200;
+  static const int created = 201;
+  static const int accepted = 202;
+  static const int nonAuthoritativeInformation = 203;
+  static const int noContent = 204;
+  static const int resetContent = 205;
+  static const int partialContent = 206;
+  static const int multiStatus = 207;
+  static const int alreadyReported = 208;
+  static const int imUsed = 226;
+  static const int multipleChoices = 300;
+  static const int movedPermanently = 301;
+  static const int found = 302;
+  static const int movedTemporarily = 302; // Common alias for found.
+  static const int seeOther = 303;
+  static const int notModified = 304;
+  static const int useProxy = 305;
+  static const int temporaryRedirect = 307;
+  static const int permanentRedirect = 308;
+  static const int badRequest = 400;
+  static const int unauthorized = 401;
+  static const int paymentRequired = 402;
+  static const int forbidden = 403;
+  static const int notFound = 404;
+  static const int methodNotAllowed = 405;
+  static const int notAcceptable = 406;
+  static const int proxyAuthenticationRequired = 407;
+  static const int requestTimeout = 408;
+  static const int conflict = 409;
+  static const int gone = 410;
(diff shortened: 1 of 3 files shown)


rutvik110:
@srujzs Regarding the naming of this api, I believe keeping it's name same as the existing `HttpStatus` api inside dart sdk shouldn't be a much of an problem, mainly due to following reasons. First is the backwards compatibility for users, second is the naming conflicts arising from `dart:io` shouldn't be a problem as the io and web library can't coexists.

lib/src/helpers/enums.dart
```diff
@@ -227,3 +227,71 @@ abstract final class Device {
   static final String propertyPrefix =
       isFirefox ? 'moz' : (isIE ? 'ms' : (isOpera ? 'o' : 'webkit'));
 }
+
+/// HTTP status codes.
```
srujzs, line 231: nit: I'd add a non-dartdoc comment here briefly mentioning that this is a copy of the `dart:_internal` version that's only exposed through `dart:html` and `dart:io`.

srujzs:
Makes sense to keep the name. LGTM, thanks!

srujzs (approved):

kevmoo:
We should bump the version and the changelog, please!

kevmoo:
Should be `1.1.0-wip`

lib/src/helpers/enums.dart
rutvik110: Updated.

rutvik110:
@srujzs @kevmoo New updates: 

* Bumped the version to `1.1.0-wip` and updated changelog.
* `HttpStatus` docs updated to add a comment requested by @srujzs.

srujzs (approved):
LGTM!

kevmoo:
thanks @rutvik110 !!

rutvik110:
Happy to contribute!!💙🎯

Issue #145, opened 2024-01-22
consider a versioning strategy for package:web


devoncarew:
`package:web` sits in an interesting design space:

- the API is generated from the web DOM IDLs
- it has a very large API surface area - around 1000 classes defined in hundreds of libraries
- is in a Dart package, versioned using semver; tldr:
  - api additions go out under minor version revs
  - breaking api changes go out under major version revs
- we want the package to reasonably closely follow the evolving set of web DOM APIs
- virtually any update to the IDLs (currently at `@webref/idl` `3.39.1`) will remove APIs, add APIs, or change signatures in ways that we would consider breaking

If we just rev. to new IDLs monthly, and follow semver strictly, we'd likely be publishing a new major version of the package every month. That frequency of major version updates _could_ cause issues for the ecosystem - if this package is in a diamond constraint for a solve, and two different package versions are being referred to.

Some ideas for how to treat this:

1. decide that monthly major releases are not an issue
2. only update the IDLs quarterly; update the semver based on changes (=> likely 4 major version revs a year)
3. decide that we will not follow semver for this package; version solves would be easier for users, but any publish could break people's CIs
4. choose a more narrow range of 'core' APIs that we make semver guarantees about; changes there would follow semver; APIs outside that core set would not have guarantees
5. break the package into n small packages based on API domain; each small package would follow semver (and would presumably need fewer overall breaking change / major version releases)
6. introduce a set of customer tests / large web app tests. apis change which break those tests would imply a new major semver release. This would be similar to 4), but with customer tests helping to figure out which APIs we cared about

cc @sigmundch @kevmoo @srujzs for thoughts

sigmundch:
I have a leaning towards (4) and (6). I'd like to use a good test suite to validate our assumptions, but I'd also like to properly communicate the policy in a way that can be understood by the community without having to scan through a corpus of tests. Also, I expect for (6) we'd also use internal apps that are not visible to the community.

Regarding (4):
* can we select a portion of the IDL that is critical to many (e.g. html elements) and is pretty stable over time?
* can we communicate to developers how to depend on the package (e.g. recommended version constraint) depending on what they are consuming?
* should we divide `web.dart` into sub-libraries depending on the version policy? (e.g. `package:web/stable.dart` vs `package:web/unstable.dart`)
* can we build tooling to validate that our policy is working well? (e.g. pub_crawl + analyzer to test how many packages break over time?)

srujzs:
[...]

PR #288, opened 2024-08-23
Use mono_repo for package:web

srujzs:
Splits package:web into two packages:
- package:web
- package:web_generator (not published)

The latter contains our tooling to generate the former, and used to be the tool directory.

- Refactors code to point to new paths (and using `Platform.script.resolve` so that the `bin/` files can be run from anywhere instead of within that directory)
- Moves directories like `third_party` and tests to their new appropriate locations
- Moves CI configurations to `mono_repo.yaml` and `mono_pkg.yaml` files and is now generated as `dart.yml` (used to be `build.yaml`)
- Updates READMEs to reflect the package split
- Adds new LICENSE, `pubspec.yaml`, and CHANGELOG files
- Removes `JSArrayExtension` which is no longer needed.

Refactors code to point to new paths and moves directories like third_party to their appropriate locations. Moves the CI configurations to mono_repo.yaml and mono_pkg.yaml files, and is now generated as dart.yml.


kevmoo:
But...why? What does this get us?

srujzs:
This came up with [analysis of an external PR](https://github.com/dart-lang/web/actions/runs/10513688196/job/29133568180?pr=286). Analysis with the latest dev version has started failing due to the generator using `JSArray.length` and `JSArray.[]`. Since those members didn't exist until recently in `dart:js_interop`, we were using the declarations in [`util.dart`](https://github.com/dart-lang/web/blob/01e861efc5547cbc7802d3262ec1a99037d4b57a/tool/generator/util.dart#L14). Once those members were added to `dart:js_interop`, it started using the definitions in the SDK instead implicitly because type members have higher precedence than extension members. However, when we added `@Since` annotations, analysis started to fail because the minimum SDK version of this package was not `>=3.6`.

The solutions were:
- Update the minimum SDK constraint. This may be okay, but it doesn't make sense to require a later SDK version even though none of the code users can actually use is impacted. This probably also isn't the last time we'll need to do this.
- Disable analysis for the generator (temporary, but not tractable long term).
- Use extension overrides for every single usage of those `JSArray` members (bad considering there are like ~25 usages).
- Separate the generation code from the interop definitions (this PR).

As a bonus, if we wanted, we could publish the generator as a separate package so users can use it to generate definitions that we don't emit in this package. This was discussed in https://github.com/dart-lang/web/issues/145 before, but is not really the primary goal of this PR.

kevmoo (approved):

README.md
```diff
@@ -1,76 +1 @@
...
-  emit code that is standards track and is not experimental to reduce the number
-  of breaking changes.
-
-## Generation and updating the package
-
-Most of the APIs in this package are generated from public assets. See
-[tool/README.md](https://github.com/dart-lang/web/tree/main/tool) for
-information on the spec and IDL versions the package was generated from, and for
-the process for updating the package.
+web/README.md
```
sigmundch, line 1: did you mean to add a link?

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2023, the Dart project authors.
+Copyright 2024, the Dart project authors.
```
sigmundch, line 1: nit: I think it's OK to keep this one as 2023

web_generator/third_party/mdn/LICENSE
sigmundch, line 1: I think we should undo this part of the change - I believe that we are still supposed to keep `third_party` at the top level of the git repository.

srujzs: Yup, it looks like it's supposed to be in the root directory. Done, thanks!

LICENSE
srujzs: Done.

README.md
srujzs: Yes :) Changed it to be a symlink.

sigmundch (approved):
Thanks!

Issue #247, opened 2024-06-01
HTMLIFrameElement contentWindow property in CORS case - cannot use postMessage()


slavap:
I am trying to migrate from html package, and figured out that it is impossible to call postMessage() for iframe with content from another domain.

```dart
myIFrame.contentWindow?.postMessage('aaa'.toJS, '*'.toJS);
```

The same code with html package works without any security exceptions.

I'm getting the following error when accessing myIFrame.contentWindow:
SecurityError: Failed to read a named property from 'Window': Blocked a frame with origin "http://localhost:51005" from accessing a cross-origin frame.

**There is a quite ugly workaround for this issue.**

Add to index.html:
```html
<script>
    window.jsMyx = {
      postMessage(iframeElt, msg) {
        if (iframeElt) {
          iframeElt.contentWindow.postMessage(msg, '*');
        }
      }
    };
  </script>
```

Then in Dart:
```dart
import 'dart:js_interop';

import 'package:web/web.dart' as web;

extension type JsMyx._(JSObject _) implements JSObject {
  external void postMessage(web.HTMLIFrameElement iframe, JSAny? msg);
}

@JS()
external JsMyx get jsMyx;
```

And the following call works without any errors again:
```dart
void _sendMessage(Map<String, String> msg) {
    if (kIsWeb) {
      HTMLIFrameElement? fr = getMyIframe();
      if (fr != null) {
        jsMyx.postMessage(fr, msg.jsify());
      }
  }
}
```

srujzs:
I suspect this is the same issue as https://github.com/dart-lang/sdk/issues/54443.

The general complication is that null-checks violate cross-origin policy as it's a `toString` call in JS. This is true for the JS compilers, whereas I believe dart2wasm likely avoids this issue due to how it does its null-checks and then immediately boxes. Using a conditional import could probably also work as a workaround if so.

Fixing this will likely involve a runtime type that dart2js/ddc knows not to call any members on, or providing some other interface that allows users to make cross-origin calls while avoiding null and type-checks in the JS compilers. `dart:html` does the latter.

slavap:
@srujzs Looks like the same problem in devtools was "solved" by disabling dart2js optimizations :-(
Also trick with safelyPostMessage() extension is not working.

PR #291, opened 2024-08-24
Add cross-origin window and location wrappers

srujzs:
Closes https://github.com/dart-lang/sdk/issues/54443
Closes https://github.com/dart-lang/web/issues/247
Closes https://github.com/dart-lang/sdk/issues/54938

Since cross-origin objects have limitations around access, wrappers are introduced to do the only safe operations. Extension methods are added to get instances of these wrappers.

I'll work on getting a test in the SDK doing the interop operations with `JSAny?` so that ability we're relying on here won't regress and we can get better coverage.

### DIFF
--- a/web/lib/src/helpers/cross_origin.dart
+++ b/web/lib/src/helpers/cross_origin.dart
@@ -1,0 +1,184 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:js_interop';
+
+import '../dom.dart' show HTMLIFrameElement, Location, Window;
+
+// Includes all the allowed and necessary APIs from the W3 spec located here:
+// https://html.spec.whatwg.org/multipage/nav-history-apis.html#cross-origin-objects
+// Some browsers may provide more access. These APIs are exposed as
+//`_DOMWindowCrossFrame` and `_LocationCrossFrame` in `dart:html`.
+
+// The Dart runtime does not allow this to be typed as any better than `JSAny?`.
+extension type _CrossOriginWindow(JSAny? any) {
+  external bool get closed;
+  external int get length;
+  // While you can set the location to a string value, this is the same as
+  // `location.href`, so we only allow the getter to avoid a
+  // `getter_not_subtype_setter_types` error.
+  external JSAny? get location;
+  external JSAny? get opener;
+  external JSAny? get parent;
+  external JSAny? get top;
+  // `frames`, `self`, and `window` are all supported for cross-origin windows,
+  // but simply return the calling window, so there's no use in supporting them
+  // for interop.
+  external void blur();
+  external void close();
+  external void focus();
+  external void postMessage(
+    JSAny? message, [
+    JSAny optionsOrTargetOrigin,
+    JSArray<JSObject> transfer,
+  ]);
+}
+
+// The Dart runtime does not allow this to be typed as any better than `JSAny?`.
+extension type _CrossOriginLocation(JSAny? any) {
+  external void replace(String url);
+  external set href(String value);
+}
+
+class CrossOriginWindow {
+  CrossOriginWindow._(JSAny? o) : _window = _CrossOriginWindow(o);
+
+  static CrossOriginWindow? _create(JSAny? o) {
+    if (o == null) return null;
+    return CrossOriginWindow._(o);
+  }
+
+  final _CrossOriginWindow _window;
+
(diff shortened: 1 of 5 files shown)


web/lib/src/helpers/cross_origin.dart
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /// The unsafe window value that this wrapper wraps that should only ever be
+  /// typed as <code>[JSAny]?</code>.
+  ///
+  /// > [!NOTE]
+  /// > This is only intended to be passed to an interop member that expects a
+  /// > <code>[JSAny]?</code>. Safety for any other operations is not
+  /// > guaranteed.
+  JSAny? get window => _window.any;
```
srujzs, line 110: We *may* not want to expose this as it may lead to issues, but I also think it's really useful to be able to pass this to interop.

srujzs: Another alternative here is providing an opaque extension type so users don't use any extension methods intended for a non-cross-origin `JSAny?` or null-assertions e.g.

```dart
extension type OpaqueCrossOriginWindow(JSAny? _) {}
```

The downside is that it can't be passed to interop methods that expect a `JSAny?`.

web/test/helpers_test.dart
```diff
@@ -36,4 +39,65 @@ void main() {
     // Ensure accessing any arbitrary item in the list does not throw.
     expect(() => dartList[0], returnsNormally);
   });
+
+  test('cross-origin windows and locations can be accessed safely', () {
```
sigmundch, line 43: (optional) I wonder if it would be worth to also add test coverage of the unwrapped behavior?  That is, something that verifies that cross-origin windows/locations fail if you tried to access them directly?

web/lib/src/helpers/cross_origin.dart
sigmundch: mmm - I think it's OK as you have it here. In the rare cases when you want this, you want to quickly pass it to JS-interop and chances are high that those interop methods expect something broad like `JSAny?`

One option here could be to rename this to something that is more indicative of the risk and generally discouraging, like `windowUnsafe`, `wrappedWindowReference`, or `unsafeWrappedWindowReference`?

web/lib/src/helpers/cross_origin.dart
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:js_interop';
+
+import '../dom.dart' show HTMLIFrameElement, Location, Window;
+
+// Includes all the allowed and necessary APIs from the W3 spec located here:
```
sigmundch, line 9: (optional) consider turning it into a dart doc to the extension type declaration below?

web/lib/src/helpers/cross_origin.dart
```diff
@@ -0,0 +1,184 @@
...
+  ]);
+}
+
+// The Dart runtime does not allow this to be typed as any better than `JSAny?`.
+extension type _CrossOriginLocation(JSAny? any) {
+  external void replace(String url);
+  external set href(String value);
+}
+
+class CrossOriginWindow {
```
sigmundch, line 44: It may be worth adding a Dart doc here like you did with CrossOriginLocation

web/lib/src/helpers/cross_origin.dart
```diff
@@ -0,0 +1,184 @@
...
+  /// > <code>[JSAny]?</code>. Safety for any other operations is not
+  /// > guaranteed.
+  JSAny? get window => _window.any;
+}
+
+/// A safe wrapper for a cross-origin location obtained through a cross-origin
+/// window.
+///
+/// Since cross-origin access is limited by the browser, the Dart runtime can't
+/// provide a type for or null-assert this value. To safely interact with this
```
sigmundch, line 117: nit: consider replacing both occurrences of "this value" in this paragraph. There is a small ambiguity that it may be misinterpreted as "instances of this class". One suggestion would be to spell out what the value is (e.g. "cross-origin objects").

sigmundch (approved):
LGTM!

kevmoo:
Also need to rebase on latest...although it looks like we're going for a 1.1 release soon.

web/test/helpers_test.dart
srujzs: I think both the test runner and `package:test` have this issue. After a good amount of debugging, I can't tell how to enable same-origin policy. Filed https://github.com/dart-lang/test/issues/2282 for now and single-stepped to verify that there are no cross-origin issues with the current test.

Issue #127, opened 2023-12-20
Add indexing operators for `getter` and `setter` attributes


parlough:
[`DOMStringMap`](https://pub.dev/documentation/web/latest/web/DOMStringMap-extension-type.html), primarily returned by [`HTMLElement#dataset`](https://pub.dev/documentation/web/latest/web/HTMLElement/dataset.html) does not provide any mechanism to set or access its values.

Would be nice if it implemented something like `Map<String, String>` if that was possible :) 

See https://github.com/dart-lang/web/blob/e89fe49d8a86845e49686b4578c56b67bdd7ba49/tool/generator/translator.dart#L492

kevmoo:
I've ran into this, too! CC @srujzs

srujzs:
I'm surprised we don't emit some properties already:

https://github.com/w3c/webref/blob/8d7678e1740d00c3b79ebf70013d0448b23e3c8b/ed/idl/html.idl#L161

It might be due to us not handling `getter` and `setter`? Maybe we should emit them using `[]` and `[]=`. https://webidl.spec.whatwg.org/#idl-special-operations

That being said, I don't think we can feasibly implement `Map` (or any Dart collections type), but hopefully the indexing operations are enough!

adrianlmm:
This one is a show stopper for me, I can't start the migration from 'dart:html' to 'web' until it gets fixed.

srujzs:
Can you describe a bit more on what's affecting your migration? If it's the above indexing operators, you can define your own `external` extension members on any interop type if they're not available: https://dart.dev/interop/js-interop/usage#extension-members-on-interop-types e.g.

```dart
import 'dart:js_interop';
import 'dart:js_interop_unsafe';

extension on DOMStringMap {
  external String operator [](String name);
  external void operator []=(String name, String value);
  void delete(String name) => this.delete(name.toJS);
}
```

Matching the definition here: https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes:domstringmap-3.

kevmoo:
![image](https://github.com/user-attachments/assets/c0ada795-9bac-4ebb-93bd-6daded367bd6)

From a grep of the IDL directory, there are a LOT of these!

rutvik110:
[...]

PR #293, opened 2024-08-27
Add [] and []= operators for unnamed getters and setters

srujzs:
Closes https://github.com/dart-lang/web/issues/127

The IDL defines unnamed getters or setters which should be emitted as special variants of the index operators.

### DIFF
--- a/web/CHANGELOG.md
+++ b/web/CHANGELOG.md
@@ -5,6 +5,8 @@
   `dart:io` and `dart:html`.
 - Added `JSImmutableListWrapper` which helps create a dart list from a JS list.
 - Deprecated `TouchListWrapper` and `TouchListConvert` in favor of `JSImmutableListWrapper`.
+- Added `[]` and `[]=` overloaded operators to types which define unnamed
+  `getter`s and `setter`s, respectively.
 
 ## 1.0.0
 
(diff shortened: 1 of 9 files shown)


srujzs:
cc @rutvik110 who was curious about generator changes. In general, changes go into `translator.dart`, which contain the following:

- Intermediate representations when translating from the parsed IDL to the resulting Dart code.
- Code to visit the parsed IDL and create IR nodes, including interfacelikes, attributes, operations, types, and more.
-- This is where the real bulk is, especially when dealing with types.
-- Here, we use the `special` field of IDL operations to give different names to some operations.
- Code to process the resulting IR nodes and emit them as Dart code (using `package:code_builder` which itself is a representation of Dart code).
-- Here, we just make sure that `[]=` members don't contain an invalid return type when emitting them as Dart code.

The definition for the parsed IDL comes from https://github.com/w3c/webidl2.js.

kevmoo:
Since we're using..."weird" operators, would it make sense to add a simple integration test...for just one of the types here.?

srujzs:
> Since we're using..."weird" operators, would it make sense to add a simple integration test...for just one of the types here.?

Yup, I tested this locally anyways, so no reason not to add a test for it. I added a smoke test for examples of `[]` and `[]=`.

kevmoo (approved):

Issue #285, opened 2024-08-20
Make constants that have a value in the IDL non-external


srujzs:
A number of APIs have constant values e.g. [`XMLHttpRequest`](https://github.com/dart-lang/web/blob/fd8c320f860ebee92947c54a3d7045bb3e6399de/lib/src/dom/xhr.dart#L94). We currently emit them as `external` values, but we can emit them as Dart members since the values are known in the IDL e.g. [`XMLHttpRequest`](https://github.com/w3c/webref/blob/e92a26a05c4e1772370239ef26b26758db720337/ed/idl/xhr.idl#L39). The only downside here is that we're ignoring what the browser implements, but in practice, I don't expect that to be different from the IDL values. The benefit of doing this is to be able to use these values in a switch-case.

kevmoo:
Didn't we discuss this somewhere else, too?

kevmoo:
I agree!

srujzs:
> Didn't we discuss this somewhere else, too?

Yup! You brought this up somewhere in an external code review but searching my mail led me to no results.

PR #294, opened 2024-08-28
Emit constants as Dart constants instead of external getters

srujzs:
Closes https://github.com/dart-lang/web/issues/285

These constants have a predetermined value according to the IDL and therefore do not need to be external. This allows users to switch over constants instead of equality checks.

Because they are now non-external, this triggers a 'constant_identifier_names' lint. To avoid ignoring this lint globally, analysis_options.yaml is forked into each package. Now unneeded ignores are pruned out of each one.


kevmoo (approved):

kevmoo:
Sweet!

PR #298, opened 2024-09-02
Add `responseHeaders` extension on `XMLHttpRequest`

mosuem:
Move this extension from [`package:http`](https://github.com/dart-lang/http/blob/b97b8dc22ea808c4fbd63f73abd7af8ecf694323/pkgs/http/lib/src/browser_client.dart#L124) here, it might also be used in [`package:grpc`](https://github.com/grpc/grpc-dart/pull/730/files#r1722304772).

---

<details>

</details>

### DIFF
--- a/web/lib/src/helpers/extensions.dart
+++ b/web/lib/src/helpers/extensions.dart
@@ -103,3 +103,30 @@ extension TouchListConvert on TouchList {
   @Deprecated('Use JSImmutableListWrapper<TouchList, Touch> directly instead.')
   List<Touch> toList() => JSImmutableListWrapper<TouchList, Touch>(this);
 }
+
+extension XMLHttpRequestGlue on XMLHttpRequest {
+  Map<String, String> get responseHeaders {
+    // from Closure's goog.net.Xhrio.getResponseHeaders.
+    final headers = <String, String>{};
+    final headersString = getAllResponseHeaders();
+    final headersList = headersString.split('\r\n');
+    for (final header in headersList) {
+      if (header.isEmpty) {
+        continue;
+      }
+
+      final splitIdx = header.indexOf(': ');
+      if (splitIdx == -1) {
+        continue;
+      }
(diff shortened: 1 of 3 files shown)


web/lib/src/helpers/extensions.dart
```diff
@@ -103,3 +103,30 @@ extension TouchListConvert on TouchList {
   @Deprecated('Use JSImmutableListWrapper<TouchList, Touch> directly instead.')
   List<Touch> toList() => JSImmutableListWrapper<TouchList, Touch>(this);
 }
+
+extension XMLHttpRequestGlue on XMLHttpRequest {
+  Map<String, String> get responseHeaders {
```
srujzs, line 108: Can we add a dartdoc comment here explaining what the output will look like? It looks like it's documented internally, so it could be as simple as reformatting that documentation. Or we could copy the `dart:html` definition:

```dart
  /// Returns all response headers as a key-value map.
  ///
  /// Multiple values for the same header key can be combined into one,
  /// separated by a comma and a space.
  ///
  /// See: http://www.w3.org/TR/XMLHttpRequest/#the-getresponseheader()-method
```

srujzs (commented):
Ideally, a simple test would be nice so this method doesn't change accidentally! You can add a test to `web/test/helpers_test.dart`.

web/lib/src/helpers/extensions.dart
```diff
@@ -103,3 +103,30 @@ extension TouchListConvert on TouchList {
...
+    // from Closure's goog.net.Xhrio.getResponseHeaders.
+    final headers = <String, String>{};
+    final headersString = getAllResponseHeaders();
+    final headersList = headersString.split('\r\n');
```
kevmoo, line 112: You could use `ListSplitter.split` here. And get a lazy iterator...

mosuem: I can't find a reference to `ListSplitter` anywhere - could you provide a link?

Issue #287, opened 2024-08-21
WebGPU generation is missing.


munrocket:
As I understand `webgpu.idl` is supported in `package:web` but it is impossible to use because `navigator.gpu` on page does not exist. Typical WebGPU example in JS starting like this:

```js
const adapter = await navigator.gpu?.requestAdapter();
const device = await adapter?.requestDevice();
if (!device) {
  fail('need a browser that supports WebGPU');
  return;
}

// Get a WebGPU context from the canvas and configure it
const canvas = document.querySelector('canvas');
const context = canvas.getContext('webgpu');
const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
context.configure({
  device,
  format: presentationFormat,
});
```
So this should compile in dart without commented lines.
```dart
import 'package:web/web.dart';
void main() async {
  console.log(window.navigator as dynamic);
  //console.log(window.navigator.gpu as dynamic);
}
```

srujzs:
[gpu](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/gpu) is still experimental alas, and we [avoid emitting such members](https://github.com/dart-lang/web/tree/main/web_generator#compatibility). Supporting this within your own application is still possible through interop though e.g.

```dart
extension on Navigator {
  external GPU get gpu;
}

extension type GPU._(JSObject _) implements JSObject {
  external JSPromise<GPUAdapter> requestAdapter(JSObject options);
  // other members
}

extension type GPUAdapter._(JSObject _) implements JSObject {
  // other members
}
```

munrocket:
Thanks.

munrocket:
I will reopen issue, because webgpu.idl source seems not full. It’s not just missing `navigator.gpu` that can be easily patched.

Actual:
https://github.com/dart-lang/web/blob/3619fd8391b32ddde18a1ca441b288ff3c39f524/web/lib/src/dom/webgpu.dart#L4

Expected (roughly):
https://github.com/gpuweb/types/blob/main/dist/index.d.ts

Expected IDL source:
https://hg.mozilla.org/mozilla-central/file/tip/dom/webidl/WebGPU.webidl

WebGPU is really hot technology and implemented in Chrome Stable and behind the flag in Safari and Firefox.

munrocket:
> Supporting this within your own application is still possible through interop though e.g.

Maybe better to use generator from this package. Too much to write by hand.

srujzs:
[...]

PR #302, opened 2024-09-12
 Add a generate-all flag to emit all bindings

srujzs:
In the case where a user wants to use an experimental API
    that we don't include in package:web, it is useful to have the
    generator emit bindings still so they can copy it over instead
    of having to write all of it themselves. This also allows us to
    find problems in our tools with newer bindings before they
    become standard.

    - Adds a generate-all flag to the tools and refactors code to
    incorporate it when determining whether to generate declarations.
    - Cleans up argument passing so we're always using package:args.
    - Adds a generate_all_and_analyze job to the Dart CI.

### DIFF
--- a/web_generator/lib/src/bcd.dart
+++ b/web_generator/lib/src/bcd.dart
@@ -15,12 +15,18 @@ import 'filesystem_api.dart';
 class BrowserCompatData {
   static final Map<String, Set<BCDPropertyStatus>> _eventHandlers = {};
 
+  /// Whether to generate all the bindings regardless of property status.
+  static bool generateAll = false;
+
   /// Returns whether [name] is an event handler that is supported in any
   /// interface.
   static bool isEventHandlerSupported(String name) =>
+      generateAll ||
       _eventHandlers[name]?.any((bcd) => bcd.shouldGenerate) == true;
 
-  static BrowserCompatData read() {
+  static BrowserCompatData read({required bool generateAll}) {
+    BrowserCompatData.generateAll = generateAll;
+
     final path =
         p.join('node_modules', '@mdn', 'browser-compat-data', 'data.json');
     final content = (fs.readFileSync(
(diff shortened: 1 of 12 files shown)


srujzs:
This came up a couple of times, but most recently in https://github.com/dart-lang/web/issues/287.

web_generator/lib/src/bcd.dart
```diff
@@ -15,12 +15,18 @@ import 'filesystem_api.dart';
 class BrowserCompatData {
   static final Map<String, Set<BCDPropertyStatus>> _eventHandlers = {};
 
+  /// Whether to generate all the bindings regardless of property status.
+  static bool generateAll = false;
```
sigmundch, line 19: I was trying to see if we could get away without making this static and without too much refactoring, and it appears that maybe it's possible?

Seems we use this property in:
* BCDItem subclasses's `shouldGenerate` getters: those however appear to be used within the context of a method that already checks for generateAll (isEventHandlerSupported and shouldGenerateInterface) so it may be safe to remove the use from these getters?
* the methods where generateAll is used also have access to a translator, so they can fetch the value through it (e.g. `_shouldGenerateMember` can read it from `Translator.instance!.browserCompatibilityData.generateAll`?

Would that work?

web_generator/lib/src/bcd.dart
```diff
@@ -84,7 +90,7 @@ class BrowserCompatData {
   BCDInterfaceStatus? retrieveInterfaceFor(String name) => interfaces[name];
 
   bool shouldGenerateInterface(String name) =>
-      retrieveInterfaceFor(name)?.shouldGenerate ?? false;
+      generateAll || (retrieveInterfaceFor(name)?.shouldGenerate ?? false);
```
sigmundch, line 93: if we can't remove the uses of `generateAll` from BCDInterfaceStatus/BCDPropertyStatus, maybe we could remove this one and the one in `isEventHandlerSupported` (seems that we'll get `true` from the underlying values, except when there isn't a interface for name (e.g. null result?)

sigmundch (approved):
LGTM, just a small suggestion below

web_generator/lib/src/bcd.dart
srujzs: Good idea! It took a little bit of refactoring, but making `shouldGenerate` a field allows this to work for the first bullet-point and the second bullet point works for translator.dart. `isEventHandlerSupported` doesn't need to check for `generateAll` since `shouldGenerate` handles that anyways.

PR #304, opened 2024-09-14
Refactor generation logic so partials come after

srujzs:
Currently, the generator adds members to interfacelikes as we process various IDL files. This may lead to partial interfaces being added before the interface definition. Instead, this CL adds partials after the interface definition, and then applies mixins. It also dedups some logic around browser compatibility so that we only check compatibility and process in one location.

While there is a diff in the generated files, it is functionally the same because all we do is move members around in the same extension type.

### DIFF
--- a/web_generator/lib/src/translator.dart
+++ b/web_generator/lib/src/translator.dart
@@ -26,9 +26,7 @@ class _Library {
   final String url;
   // Contains both IDL `interface`s and `namespace`s.
   final List<idl.Interfacelike> interfacelikes = [];
-  final List<idl.Interfacelike> partialInterfaces = [];
   final List<idl.Interfacelike> interfaceMixins = [];
-  final List<idl.Interfacelike> partialInterfaceMixins = [];
   final List<idl.Typedef> typedefs = [];
   final List<idl.Enum> enums = [];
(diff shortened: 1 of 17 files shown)


web_generator/lib/src/translator.dart
```diff
@@ -732,34 +714,41 @@ class Translator {
...
+      _addOrUpdateInterfaceLike(interfacelike);
+    }
+    final mixins = _includes[interfacelikeName];
+    if (mixins == null) return;
+    for (final mixin in mixins) {
       for (final interfacelike in [
-        ...library.interfaceMixins,
-        ...library.partialInterfaceMixins
+        _typeToDeclaration[mixin] as idl.Interfacelike,
+        ..._typeToPartials[mixin] ?? <idl.Interfacelike>[]
```
sigmundch, line 738: nit: use `...?_typeToPartials[mixin]` instead of `??`?

sigmundch (approved):

web_generator/lib/src/translator.dart
srujzs: Ooh, I don't think I knew this existed. Thanks!

PR #305, opened 2024-09-21
onLoad helper for window and document

isoos:
Also updated the order of the fields in `WindowEventGetters`.


srujzs (approved):
LGTM, thanks!

PR #345, opened 2025-02-25
Update APIs and docs

kevmoo:


srujzs:
Looks like `catch` and `finally` are not in our reserved keywords list. I'll upload a change handling that.

kevmoo:
Fixed!

srujzs (approved):

PR #359, opened 2025-04-21
Migrate web_generator to new element model

srujzs:
New version of analyzer requires a migration: https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/element_model_migration_guide.md. Makes Dart CI green again.


kevmoo (changes requested):
You'll want to bump the min pkg:analyzer version in the pubspec, too!

srujzs:
> You'll want to bump the min pkg:analyzer version in the pubspec, too!

Ah right, the pubspec didn't change, so I guess the analyzer version must have just resolved to a newer version. Done!

kevmoo (approved):

PR #368, opened 2025-06-03
[web_generator] Setting up `web_generator` for Dart JS Interop Gen

nikeokoronkwo:
This PR is for setting up the `web_generator` to begin adding support for generating Dart JS interop interface code from TypeScript Declaration Code.

For more information, see the [project overview](https://github.com/dart-lang/sdk/blob/main/docs/gsoc/Dart-GSoC-2025-Project-Ideas.md#idea-generate-js-interop-bindings-for-typescript-declaration-files)

<details>

</details>

### DIFF
--- a/web_generator/README.md
+++ b/web_generator/README.md
@@ -90,10 +90,10 @@ experimental and non-standard APIs.
 
 Based on:
 
-<!-- START updated by bin/update_bindings.dart. Do not modify by hand -->
+<!-- START updated by bin/update_idl_bindings.dart. Do not modify by hand -->
 | Item | Version |
 | --- | --: |
 | `@webref/css` | [6.20.3](https://www.npmjs.com/package/@webref/css/v/6.20.3) |
 | `@webref/elements` | [2.4.0](https://www.npmjs.com/package/@webref/elements/v/2.4.0) |
 | `@webref/idl` | [3.60.1](https://www.npmjs.com/package/@webref/idl/v/3.60.1) |
-<!-- END updated by bin/update_bindings.dart. Do not modify by hand -->
+<!-- END updated by bin/update_idl_bindings.dart. Do not modify by hand -->


web_generator/bin/generate_dart_ts_bindings.dart
```diff
@@ -0,0 +1,104 @@
...
+    print('''
+${ansi.lightRed.wrap('At least one argument is required')}
+
+$_usage''');
+    return;
+  }
+
+  assert(p.fromUri(Platform.script).endsWith(_thisScript.toFilePath()));
+
+  // begin compiling the main dart entry
```
srujzs, line 42: We should likely move this to shared code as we're relying on the same process compiled file (`dart_main.dart`) anyways. It also seems like we're sharing the calls to `npm`, `dart compile`, and `node`, so what if the shared code just has functions to do those specific things and not functions to run any process e.g. `npmUpdate`, `npmInstall`, `compileDartMain`, `runDartMainWithNode` instead.

web_generator/lib/src/_js_supertypes_src.dart
```diff
@@ -0,0 +1,4 @@
+import 'dart:js_interop';
```
srujzs, line 1: It took me a while to figure out the purpose of this file. :D I wonder if we should instead create a temporary file for analysis that we delete after calculating the JS type supertypes. The contents could just be what we have here.

web_generator/lib/src/dart_main.dart
```diff
@@ -24,16 +26,44 @@ void main(List<String> args) async {
...
-  );
+
+  final argResult = _parser.parse(args);
+
+  switch (argResult.rest.first.toLowerCase()) {
+    case 'idl':
+      await generateIDLBindings(
+        outputDirectory: argResult['output'] as String,
+        generateAll: argResult['generate-all'] as bool,
+        languageVersion: Version.parse(languageVersionString),
```
srujzs, line 37: We'll probably need this still to format the generated files, but totally okay waiting to add this until then too.

web_generator/test/ts_bindings_test.dart
```diff
@@ -0,0 +1,75 @@
...
+  final testGenDTSFiles = p.join(testGenFolder, 'input');
+
+  final bindingsGenPath = p.join('lib', 'src');
+
+  group('Web Generator TS Bindings Integration Test', () {
+    final inputDir = Directory(testGenDTSFiles);
+
+    setUp(() async {
+      // set up npm
+      await runProcess(
```
srujzs, line 22: With the above comment to just refactor the `runProcess` calls, we could just use those `cli.dart` functions directly here instead of this.

srujzs (commented):
Very cool to see us adding TSC and potentially start using it!

web_generator/bin/generate_dart_ts_bindings.dart
nikeokoronkwo: Would it be fine if the `ArgParser` for this and the IDL are combined into one, separated by commands (i.e calls to `bin/update_bindings.dart` become `bin/update_bindings.dart idl` for IDL and `bin/update_bindings.dart dts` for Typescript Decl to Dart Interop Gen)

kevmoo:
The tests don't run by default because they are VM tests.
And when I run them locally they fail because there is not test input/output data.

kevmoo:
Also

```diff
diff --git a/web_generator/test/ts_bindings_test.dart b/web_generator/test/ts_bindings_test.dart
index 3470d7a..8ef208a 100644
--- a/web_generator/test/ts_bindings_test.dart
+++ b/web_generator/test/ts_bindings_test.dart
@@ -9,13 +9,12 @@ import 'package:test/test.dart';
 import 'package:web_generator/src/cli.dart';
 
 void main() {
-  final testGenFolder = p.join('test', 'gen');
-  final testGenDTSFiles = p.join(testGenFolder, 'input');
-
   group('Web Generator TS Bindings Integration Test', () {
+    final testGenFolder = p.join('test', 'gen');
+    final testGenDTSFiles = p.join(testGenFolder, 'input');
     final inputDir = Directory(testGenDTSFiles);
 
-    setUp(() async {
+    setUpAll(() async {
       // set up npm
       await runProc('npm', ['install'],
           workingDirectory: bindingsGeneratorPath, detached: true);
```

Should keep the group vars together.
Should use `setUpAll` so the compile is only run once – since there should be ZERO change between tests. But that's once we have tests 😄

srujzs:
> The tests don't run by default because they are VM tests.
> And when I run them locally they fail because there is not test input/output data.

It looks like we were skipping a test already because of this: https://github.com/dart-lang/web/pull/373.

@nikeokoronkwo I filed https://github.com/dart-lang/web/issues/372 to track unskipping this test now that I'm enabling it in the CI. We should add a `data` folder that we read from and have the test use that.

PR #373, opened 2025-06-06
Skip ts_bindings_test and run web_generator tests with VM

srujzs:
It looks like we were silently skipping scrape_mdn_test.dart before because we never ran the tests in this package with the VM.


kevmoo (approved):

srujzs:
Thanks!

PR #377, opened 2025-06-07
[web_generator] Add Support for Parsing IDL Files

### LINKED ISSUE #374: [web_generator] Support for Single IDL File Parsing
This issue is for supporting the parsing of individual IDL files, rather than the default of the ones provided for `package:web`. 
This will most likely involve adding support on `dart_main.dart` for individual IDL files, and default to normal behaviour if an input file isn't provided. 

This can help us run integration tests on the generation process for IDL files.

nikeokoronkwo:
Fixes #374

This Pull Request extends the functionality of IDL generation by allowing support for parsing individual IDL files and definitions.

This also contains an integration test suite `integration/idl_test.dart` for testing IDL generation against expected files in `integration/idl/`

For more information, check #374

<details>
</details>


.gitignore
```diff
@@ -5,3 +5,5 @@ pubspec.lock
 web_generator/lib/src/*.js
 web_generator/lib/src/*.js.*
 web_generator/lib/src/node_modules/
+
+_temp
```
kevmoo, line 9: Remove, please!

```suggestion
```

nikeokoronkwo: This was because the integration test involves generating files to compare
Something similar to: https://github.com/dart-lang/native/blob/main/pkgs/swift2objc/test/integration/temp/.gitignore

kevmoo: Consider putting these in `.dart_tool/[pkg_name]` directory if possible.

.gitignore
```diff
@@ -4,4 +4,4 @@ pubspec.lock
 
 web_generator/lib/src/*.js
 web_generator/lib/src/*.js.*
-web_generator/lib/src/node_modules/
+web_generator/lib/src/node_modules/
```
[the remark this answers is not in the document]
kevmoo: 😄

web_generator/test/integration/idl/types_input.idl
```diff
@@ -0,0 +1,6 @@
+interface Global {
+  const boolean DEBUG = false;
+  const octet LF = 10;
+  const unsigned long BIT_MASK = 0x0000fc00;
+  const double AVOGADRO = 6.022e23;
+};
```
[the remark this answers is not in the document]
nikeokoronkwo: My bad.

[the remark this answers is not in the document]
kevmoo: No biggy! just FYI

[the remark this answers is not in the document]
nikeokoronkwo: Thanks

web_generator/bin/update_idl_bindings.dart
```diff
@@ -67,11 +67,13 @@ $_usage''');
 
   // Run app with `node`.
   final generateAll = argResult['generate-all'] as bool;
+  final inputFiles = argResult.rest;
```
srujzs, line 70: Hmm, instead of just treating the rest of the args as files, what if we used a multi-option?

web_generator/bin/update_idl_bindings.dart
```diff
@@ -161,18 +163,13 @@ final _endComment =
     '<!-- END updated by $_scriptPOSIXPath. Do not modify by hand -->';
 
 final _usage = '''
-Global Options:
-${_parser.usage}
+${ansi.styleBold.wrap('WebIDL Gen')}:
+$_thisScript [... .idl file] [options]
```
srujzs, line 167: We can get rid of this if use a multi-option and make it explicit that if that option is not used, we generate the definitions for `package:web`.

web_generator/bin/update_idl_bindings.dart
```diff
@@ -67,11 +67,13 @@ $_usage''');
...
   // Run app with `node`.
   final generateAll = argResult['generate-all'] as bool;
+  final inputFiles = argResult.rest;
   await runProc(
     'node',
     [
       'main.mjs',
       '--idl',
+      for (String inputFile in inputFiles) '--input=$inputFile',
       '--output=${p.join(_webPackagePath, 'lib', 'src')}',
```
srujzs, line 77: If `inputFiles` is non-empty, should we require users specify an output so we don't place the files in the `package:web` destination?

web_generator/lib/src/dart_main.dart
```diff
@@ -64,22 +68,51 @@ Future<void> generateJSInteropBindings({
...
+      final libraryPath = entry.key;
+      final library = entry.value;
+
+      final contents = _emitLibrary(library, languageVersion).toJS;
+      fs.writeFileSync('$output/$libraryPath'.toJS, contents);
+    }
+  } else {
+    final single = input.length == 1;
+    // parse individual files
+    ensureDirectoryExists(single ? p.dirname(output) : output);
```
srujzs, line 95: Why do these differ when there's a single input? Shouldn't we always check the dirname of the output directory?

web_generator/lib/src/dart_main.dart
```diff
@@ -64,22 +68,51 @@ Future<void> generateJSInteropBindings({
...
+
+    final bindings = await generateBindingsForFiles({
+      for (final file in input)
+        file: fs
+            .readFileSync(file.toJS, JSReadFileOptions(encoding: 'utf-8'.toJS))
+            .dartify() as String
```
srujzs, line 101: Prefer: `(fs.readFileSync(file.toJS, JSReadFileOptions(encoding: 'utf-8'.toJS)) as JSString).toDart`

web_generator/lib/src/generate_bindings.dart
```diff
@@ -86,3 +89,20 @@ Future<TranslationResult> generateBindings(
   translator.addInterfacesAndNamespaces();
   return translator.translate();
 }
+
+Future<TranslationResult> generateBindingsForFiles(
+    Map<String, String> fileContents, String output, bool single) async {
+  final cssStyleDeclarations = await _generateCSSStyleDeclarations();
```
srujzs, line 95: nit: I think this is okay because they're necessary dependencies for some IDL files, but it naturally does feel weird to do this for when inputs are specified. I'd just add a comment saying that we need to generate the style declarations and element tag map in case they're used in the input files.

web_generator/lib/src/dart_main.dart
```diff
@@ -64,22 +68,51 @@ Future<void> generateJSInteropBindings({
...
+    }, output, single);
+
+    for (var entry in bindings.entries.where((e) => e.key != 'dom.dart')) {
+      final libraryPath = entry.key;
+      final library = entry.value;
+
+      final contents = _emitLibrary(library, languageVersion).toJS;
+      // print('FOR $libraryPath \n${'='*20}\n $contents');
```
srujzs, line 109: nit: remove.

web_generator/lib/src/dart_main.dart
```diff
@@ -64,22 +68,51 @@ Future<void> generateJSInteropBindings({
...
+      if (single) {
```
srujzs, line 110: I think we should always generate to a specified output directory if input files are specified. Assuming the user wants the output in a `lib/` folder but only in the non-single case is a bit awkward.

web_generator/lib/src/dart_main.dart
```diff
@@ -64,22 +68,51 @@ Future<void> generateJSInteropBindings({
...
+    ensureDirectoryExists(single ? p.dirname(output) : output);
+
+    final bindings = await generateBindingsForFiles({
+      for (final file in input)
+        file: fs
+            .readFileSync(file.toJS, JSReadFileOptions(encoding: 'utf-8'.toJS))
+            .dartify() as String
+    }, output, single);
+
+    for (var entry in bindings.entries.where((e) => e.key != 'dom.dart')) {
```
srujzs, line 104: This should only be produced when we generate for `package:web`, so I'd pass an option to `translate` to conditionally not generate this file instead of filtering by name later.

web_generator/test/integration/idl_test.dart
```diff
@@ -0,0 +1,68 @@
...
+
+    for (final inputFile in inputDir
+        .listSync()
+        .whereType<File>()
+        .where((f) => p.basenameWithoutExtension(f.path).endsWith('_input'))) {
+      final inputFileName = p.basenameWithoutExtension(inputFile.path);
+      final inputName = inputFileName.replaceFirst('_input', '');
+
+      final outputActualPath =
+          p.join('.dart_tool', 'idl', '${inputName}_actual.dart');
```
srujzs, line 33: Let's use a temp directory instead using `dart:io`.

web_generator/test/integration/idl_test.dart
```diff
@@ -0,0 +1,68 @@
...
+      final outputExpectedPath =
+          p.join(testGenFolder, '${inputName}_expected.dart');
```
srujzs, line 35: Did you mean to include these files in this PR? I only see the input IDL files.

srujzs (commented):
Nice, thanks Nike!

web_generator/test/integration/idl_test.dart
nikeokoronkwo: What if we might want to look at the actual output and compare?

web_generator/lib/src/dart_main.dart
srujzs: Ah right, this was in a previous change. It may be cleaner to always and only accept a directory for the output if we need to support the multiple input case anyways. Thoughts?

kevmoo:
@srujzs – running tests. Is this good?

srujzs:
We chatted a bit before about some of the failures for the expected files and I believe Nike is still planning on adding the expected outputs into this PR before we can land.

srujzs (approved):
Looking good! Just one final nit.

PR #382, opened 2025-06-11
Add Variable Declaration Support

### LINKED ISSUE #375: Interop Gen: Support Variable Declarations
This issue is to add support to the dart interop interface gen on `web_generator` for variable declarations in TypeScript declaration files

```ts
declare const myConstVar: number = 9;
declare let myMutableVar: string = "mutable";
declare var myVar: boolean = false;
```

These will be mapped to `@JS` getters and setters

nikeokoronkwo:
Fixes #375 

This pull requests aims to add support for Variable declaration to the Dart JS Interop Generation Tool, as well as introduce features to be used for supporting other declarations

### TODO
- [ ] Generate Spec(s) from Declarations
- [ ] Set up Integration Test Suite for Variable Declarations
- [ ] Finalize Setup for future support for:
  - [ ] Type Declarations (Referred types)
  - [ ] Declaration Lookup and Transform
  - [ ] Nullable Types when union with `undefined`
  - [ ] Filtering Declarations


web_generator/lib/src/ast.dart
```diff
@@ -0,0 +1,136 @@
...
+          ..external = true
+        )
+      );
+    }
+
+    return codeBlocks.first;
+  }
+}
+
+enum VariableModifier { let, $const, $var }
```
kevmoo, line 136: missing new line at the end!

kevmoo:
@srujzs ?

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -0,0 +1,214 @@
...
+    final otherDecls = filteredDeclarations.entries
+        .map((e) => _filterForDecl(e.value))
+        .reduce((value, element) => value..addAll(element));
+
+    return filteredDeclarations..addAll(otherDecls);
+  }
+
+  /// Given an already filtered declaration [decl],
+  /// filter out dependencies of [decl] and return them as a declaration map
+  DeclarationMap _filterForDecl([Node? decl]) {
```
srujzs, line 92: Maybe a slightly more descriptive name like `_getDependenciesOfDecl`. I'd also update the doc to reflect that this is recursive.

web_generator/lib/src/ast.dart
```diff
@@ -0,0 +1,170 @@
+import 'package:code_builder/code_builder.dart';
+
+import 'interop_gen/generate.dart';
+
+sealed class Node {
+  abstract final String? name;
+  abstract final String id;
```
srujzs, line 7: Should this be an `ID`? Is there any reason to use a `String` that we have to parse later versus just creating the `ID` ahead of time?

web_generator/lib/src/ast.dart
```diff
@@ -0,0 +1,170 @@
...
+      required this.type,
+      required this.modifier,
+      required this.exported});
+
+  @override
+  String get id => 'var#$name';
+
+  @override
+  List<Spec> emit() {
+    // generate a getter and setter pair
```
srujzs, line 145: Another option is to emit an external field instead (which incapsulate both a getter and a setter) to reduce the total number of members generated and potentially only emit one member per `VariableDeclaration`.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -0,0 +1,214 @@
...
+          _ => throw Exception('Unsupported Declaration Kind: ${node.kind}')
+        };
+        // ignore: dead_code This line will not be dead in future decl additions
+        declarationMap[decl.id] = decl;
+    }
+
+    nodes.add(node);
+  }
+
+  DeclarationMap filter([bool Function(Node)? saveIf]) {
```
srujzs, line 50: I'd wait until we use `saveIf` to add the logic for it to simplify the PR.

srujzs (commented):
Thanks Nike, super cool to see! I added a first pass of comments. In general, I'd go through and add Dart doc comments on non-obvious declarations to detail what the function of each method is. I'd also wait to add functionality that doesn't get used yet so this PR is more focused.

web_generator/lib/src/ast.dart
```diff
@@ -0,0 +1,170 @@
...
+    final codeBlocks = <Method>[];
+
+    codeBlocks.add(Method((m) => m
+      ..name = name
+      ..type = MethodType.getter
+      ..annotations.add(generateJSAnnotation())
```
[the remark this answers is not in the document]
nikeokoronkwo: For now, I wanted to handle variables and properties differently, until I had gotten to interfaces, where they'd probably share some code. So for now, all variables are top level.

web_generator/lib/src/interop_gen/transform.dart
```diff
@@ -0,0 +1,118 @@
...
+      }).reduce((val, element) => [...val, ...element]);
+      final lib = Library((l) => l..body.addAll(specs));
+      return MapEntry(file, formatter.format('${lib.accept(emitter)}'));
+    });
+  }
+}
+
+class TransformationOptions {
+  /// whether we are outputting to a single file or to multiple files
+  bool singleFile;
```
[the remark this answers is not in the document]
nikeokoronkwo: I was considering it for the code generation

web_generator/lib/src/ast.dart
```diff
@@ -0,0 +1,170 @@
...
+
+sealed class Node {
+  abstract final String? name;
+  abstract final String id;
+  final String? dartName;
+
+  Node() : dartName = null;
+}
+
+abstract class Declaration extends Node {
```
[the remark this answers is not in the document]
nikeokoronkwo: Do you mean anonymous types?

web_generator/lib/src/interop_gen/transform.dart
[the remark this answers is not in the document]
srujzs: Similar comment thread in the other PR, but maybe this goes away if we always expect a directory output.

web_generator/lib/src/ast.dart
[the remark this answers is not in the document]
srujzs: Makes sense, thanks for the explanation!

web_generator/lib/src/ast.dart
```diff
@@ -1,25 +1,30 @@
 import 'package:code_builder/code_builder.dart';
```
srujzs, line 1: Remember to include the license in every new Dart file:

```
// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
// for details. All rights reserved. Use of this source code is governed by a
// BSD-style license that can be found in the LICENSE file.
```

web_generator/lib/src/interop_gen/transform.dart
```diff
@@ -25,60 +25,36 @@ class TransformResult {
...
       final lib = Library((l) => l..body.addAll(specs));
       return MapEntry(file, formatter.format('${lib.accept(emitter)}'));
     });
   }
 }
 
-class TransformationOptions {
-  /// whether we are outputting to a single file or to multiple files
-  bool singleFile;
+/// A map of declarations, where the key is the declaration's ID.
```
srujzs, line 37: ubernit: `declaration's stringified ID.`

web_generator/lib/src/ast.dart
```diff
@@ -117,17 +122,14 @@ class UnionType extends Type {
 
 class VariableDeclaration extends NamedDeclaration
```
srujzs, line 123: We talked briefly about having the IDL use this AST as well, but I'm okay either if you want to do this now as you're adding nodes or later when we have the necessary AST for the IDL generation.

srujzs (approved):
LGTM % my comments, thanks!

web_generator/lib/src/ast.dart
nikeokoronkwo: I'll add this later when I'm working with nodes similar to those in the IDL. Since we didn't find any instances of global declarations, the next AST declaration would be interfaces

srujzs (approved):
Thanks Nike!

PR #386, opened 2025-06-11
Add Support for Configuration of Dart JS Interop Gen

### LINKED ISSUE #376: [web_generator] Configuration Support
This issue is to track adding basic configuration support for both the Dart Interop Generator and the IDL Generator (mainly the former, as there is more configuration being done)

The result should be a base `Config` class applicable to both the IDL and Dart Interop Generator.

The configuration for the Dart Interop Generator will be modified as declarations are added

nikeokoronkwo:
Fixes #376 

This issue adds basic support for configuration for the dart js interop generator. Configuration will be extended over time as support for different typescript declarations are added to the project.

### TODO
- [ ] Should we add config for IDL as well?


kevmoo:
note you need to rebase here!

PR #407, opened 2025-07-06
[interop] Add Support for Typealiases

### LINKED ISSUE #406: Interop Gen: Support Typealiases
This issue is to add support for typealiases in the Dart JS Interop Interface Generator for typealiases.

```ts
type TSType = string;
type TSGenType<T> = Promise<T>;
```

nikeokoronkwo:
Fixes #406

This issue adds support for TS typealises, denoted by the `type` keyword, and represents them as Dart `typedef`s. 
```ts
type A = string;
type B<T> = Promise<T>;
```
turns into
```dart
typedef A = JSString;
typedef B<T extends JSAny> = JSPromise<T>;
```


web_generator/lib/src/ast/declarations.dart
```diff
@@ -138,3 +138,40 @@ class ParameterDeclaration {
...
+    implements ExportableDeclaration {
+  @override
+  final String name;
+
+  final List<GenericType> typeParameters;
+
+  final Type type;
+
+  @override
+  final String? dartName;
```
srujzs, line 152: nit: this is unrelated to this change, but we should maybe rename this to `dartRename` at some point. It's less obvious in subtypes of `NamedDeclaration` that it's only ever used when we have to give a rename.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -91,6 +93,30 @@ class Transformer {
...
+        }) ??
+        false;
+
+    final typeParams = typealias.typeParameters?.toDart;
+
+    final type = typealias.type;
+
+    return TypeAliasDeclaration(
+        name: name,
+        // TODO: Can we find a way not to make the types be JS types
```
srujzs, line 111: Yeah, this is something similar to what I came across in the IDL.

One option is to desugar the type alias in the constraint of a generic type e.g. `T extends NumberAlias` becomes `T extends JSNumber` instead. We do this in the IDL generation whenever a typedef is used in a location where a JS type is needed, but the typedef is not a valid JS type. https://github.com/dart-lang/web/blob/fb8a14921e6a402d35f4d7b61fa2dc9fde567e64/web_generator/lib/src/translator.dart#L859

web_generator/test/integration/interop_gen/typealias_input.d.ts
```diff
@@ -0,0 +1,14 @@
+export type Username = string;
+export type Age = number;
+export type IsActive = boolean;
+export type Tags = string[];
+export type List<T> = T[];
+export type Box<T> = Array<Array<T>>;
+export type PromisedArray<U, T extends Array<U>> = Promise<T>;
```
srujzs, line 7: Let's add a test case that is relevant to the TODO in `transformer.dart:111`.

web_generator/lib/src/ast/declarations.dart
nikeokoronkwo: Yes. I can do that in another PR

web_generator/lib/src/interop_gen/transform/transformer.dart
nikeokoronkwo: Could you explain a bit more? I do not seem to understand so much, and sorry for that.
- Should the name matter that much. If it does then I didn't even consider that
- My issue here is that using typealiases in generics when their base type isn't a JS type leads to issues when having code such as `Promise<NumberAlias>` being wrong because `NumberAlias = num`. If this is what you're referring to, then, like above, I may not be understanding (so sorry for that)

web_generator/test/integration/interop_gen/typealias_input.d.ts
nikeokoronkwo: Lines 11 should suffice, but I'll add more

srujzs: Ah I failed to see `List<Username>`, thanks!

web_generator/lib/src/interop_gen/transform/transformer.dart
srujzs: > sorry for that

No problem!

> Should the name matter that much. If it does then I didn't even consider that

The naming doesn't really matter - I'm referring to the problem in your second bulletpoint. You want to make sure the type argument in a `JSPromise` or `JSArray` (or any other generic JS interop type that has a constraint to `JSAny?`) respects the bound.

> An idea I have is using getJSTypeAlternative on type params of referred types by default anyways

This was kind of what I was going for. "Desugar" the type when we need it in a generic type if the type isn't a subtype of `JSAny?`. It looks like `typeArg` is the equivalent to `onlyEmitInteropTypes` here.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -227,19 +253,24 @@ class Transformer {
   }
 
   GenericType _transformTypeParamDeclaration(
-      TSTypeParameterDeclaration typeParam) {
+      TSTypeParameterDeclaration typeParam,
+      {bool shouldUseJSConstraint = true}) {
```
srujzs, line 257: I may be missing something, but this isn't set in any of the calls. Should we remove this parameter?

web_generator/lib/src/ast/declarations.dart
srujzs: No problem, thanks!

srujzs (approved):

web_generator/lib/src/interop_gen/transform/transformer.dart
nikeokoronkwo: Yes. That way I was able to know to use this only when dealing with type args or type param constraints

srujzs (approved):

PR #408, opened 2025-07-09
[interop] Refactoring Generator to make use of configuration options

nikeokoronkwo:
This PR is to connect configuration options into the generator itself, allowing for direct configuration of the generator based on a configuration file or object.

By the end of this PR, we should have
- [x] Configuration Options for declarations supported so far
- [x] Support for Filtering Declarations (by name for now, glob could be in a future PR)
- [x] Configuring code gen via config options


web_generator/lib/src/config.dart
```diff
@@ -95,36 +115,58 @@ class YamlConfig implements Config {
...
         name: yaml['name'] as String?,
         description: yaml['description'] as String?,
         languageVersion: yaml['language_version'] as String?,
-        preamble: yaml['preamble'] as String?);
+        preamble: yaml['preamble'] as String?,
+        // TODO: Can we consider using `json_serializable`?
+        functions: FunctionConfig(
+            varArgs: (yaml['functions'] as YamlMap?)?['varargs'] as int?),
+        include: (yaml['include'] as YamlList?)
+                ?.map<String>((node) => node is String ? node : node.toString())
```
srujzs, line 168: nit: Do we need to check for this instead of just `node.toString()` always?

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -23,17 +23,29 @@ class Transformer {
...
+      }).toList();
+
   /// namer, for giving elements unique names
   final UniqueNamer namer;
 
   final ProgramDeclarationMap programMap;
 
   Transformer(this.programMap, this.typeChecker,
-      {Iterable<String> exportSet = const []})
+      {Iterable<String> exportSet = const [],
```
srujzs, line 45: nit: If we're always converting these iterables to a subtype, expect the subtype instead e.g. `Set<String> exportSet` and `List<String> filterDeclSet`. This avoids the case where we unnecessarily call e.g. `toList` on a `List`.

srujzs (approved):
LGTM!

PR #415, opened 2025-07-12
[interop] Support Classes and Interfaces

### LINKED ISSUE #402: Interop Gen: Support Interfaces
This issue is to add support to the Dart TS Declaration Interop Gen tool to support TS interfaces.

Interfaces are represented as anonymous objects (unimplemented) or are implemented by classes:
```ts
export interface A {
  p: number;
  c: string;
  control(a: number): void;
}
```

This issue will also require the refactoring of variables and functions to consider properties and methods.

### LINKED ISSUE #409: Interop Gen: Support Classes
This issue is to add support for TypeScript Classes in the Dart TS Interop Interface Generator.

```ts
class MyClass {
  constructor();
  private managed: boolean;
  // more declarations
}
```

### LINKED ISSUE #411: Interop Gen: Support Overloaded Constructors
This issue adds support for overloaded constructors to the Dart JS Interop Interface Generator.

```ts
class Example {
  constructor(value: string);
  constructor(value: number);
  constructor(value: any) {
    console.log(value);
  }
}

```

nikeokoronkwo:
Fixes #402 
Fixes #409
Fixes #411

This PR adds support for classes and interfaces in TypeScript for the Dart JS Interop Interface Generator.

Interfaces and Classes in TS are mapped to Dart extension types, following Dart's new JS Static Interop for JS. Constructors are generated by default for classes (defaulting to a zero-parameter constructor if not specified). Unless denoted by hierarchy, all extension types by default have `JSObject` as their rep type and implement `JSObject`

1. Constructors are mapped to Dart constructors on an extension type
2. Methods are mapped to Dart methods
3. Properties are mapped to Dart properties
4. `get` and `set` accessors are mapped to getters and setters in Dart
5. callable interfaces are mapped to extension types with a `call` parameter, so they can be called in Dart.
6. indexable interfaces and classes are mapped to override the operators of `[]` and `[]=` (the second is only if the indexer member is not `readonly`), so they can be indexed in Dart.
7. `readonly` members are mapped to only getters rather than normal fields
8. Only `public` members are generated, as those are the only members that can be accessed by Dart's JS Interop layer.
9. Alternate constructors are mapped to factory constructors with different names in the extension type
10. Overriden methods are marked with [`@redeclare`](https://pub.dev/documentation/meta/latest/meta/redeclare-constant.html)
11. Nullable Methods are represented as getters that return `JSFunction?` (once [dart-lang/sdk#54557](https://github.com/dart-lang/sdk/issues/54557) lands, we will have it with generics)
12. Extended classes are represented with the extendee being implemented _as well as_ being it's representation type
13. Implemented classes and extended interfaces are represented with the extendees/implementees being implemented on the extension type.

Therefore, the following
```ts
interface Shape<TMetadata = any> {
    readonly id: string;
    name: string;
    area(): number;
    area(unit: 'cm2' | 'in2'): string;
    onUpdate?(prev: Shape<TMetadata>): void;
}
declare class ShapeImpl<TMeta = any> implements Shape<TMeta> {
    readonly id: string;
    name: string;
    /* other decls in Shape */
    metadata?: TMeta;
    constructor(name: string, type?: 'circle' | 'rectangle' | 'polygon');
    constructor(config: Shape<TMeta>);
    get location(): string;
    set location(value: string);
    static getById(id: string): ShapeImpl;
    toString(): string;
}
```
Maps to this
```dart
import 'dart:js_interop' as _i1;
import 'package:meta/meta.dart' as _i2; // needed for @redeclare

extension type Shape<TMetadata extends _i1.JSAny>(_i1.JSObject _) implements _i1.JSObject {
  external String get id;
  external String name;
  external double area();
  @_i1.JS('area')
  external String area$1(AnonymousUnion unit); // AnonymousUnion defined afterwards
  external JSFunction? get onUpdate;
}

extension type ShapeImpl <TMeta extends _i1.JSAny>(_i1.JSObject _) implements Shape {
  external ShapeImpl(String name, [AnonymousUnion$1 type]);
  external factory ShapeImpl.unnamed$1(Shape<TMeta> config);
  @_i2.redeclare
  external String get id;
  external String name;

  external TMeta? metadata;
  external String get location;
  external set location(String value);
  external static ShapeImpl getById(String id)
  /* other decls in Shape */
}
```

For more examples, check out the added tests `classes_input.d.ts` and `interfaces_input.d.ts` as well as their expected results for more

Other features, such as an adjusted AST, filtering mechanism, and more have been implemented.

> This PR also adds support for a `typeMap` to allow reusage of certain types with associated declarations such as unions

### DIFF
--- a/web_generator/lib/src/ast/declarations.dart
+++ b/web_generator/lib/src/ast/declarations.dart
@@ -6,18 +6,108 @@ import 'package:code_builder/code_builder.dart';
 
 import '../interop_gen/namer.dart';
 import 'base.dart';
+import 'builtin.dart';
 import 'helpers.dart';
 import 'types.dart';
 
+/// A declaration that defines a type
+///
+// TODO: Add support for `ClassOrInterfaceDeclaration`
+//  once implementing namespaces and module support
 sealed class TypeDeclaration extends NamedDeclaration
     implements ExportableDeclaration {
-  abstract final List<GenericType> typeParameters;
+  @override
+  final String name;
 
-  abstract final List<MethodDeclaration> methods;
+  @override
+  final String? dartName;
 
-  abstract final List<PropertyDeclaration> properties;
+  @override
+  final bool exported;
 
-  abstract final List<OperatorDeclaration> operators;
+  final List<GenericType> typeParameters;
+
+  final List<MethodDeclaration> methods;
+
+  final List<PropertyDeclaration> properties;
+
+  final List<OperatorDeclaration> operators;
+
+  final List<ConstructorDeclaration> constructors;
+
+  TypeDeclaration(
+      {required this.name,
+      this.dartName,
+      required this.exported,
+      this.typeParameters = const [],
+      this.methods = const [],
+      this.properties = const [],
+      this.operators = const [],
+      this.constructors = const []});
+
+  ExtensionType _emit(
+      [covariant DeclarationOptions? options,
+      bool abstract = false,
+      List<Type> extendees = const [],
+      List<Type> implementees = const []]) {
+    options ??= DeclarationOptions();
+
+    final hierarchy = getMemberHierarchy(this);
+
+    final fieldDecs = <Field>[];
+    final methodDecs = <Method>[];
+
+    bool isOverride(String name) =>
+        hierarchy.contains(name) && GlobalOptions.redeclareOverrides;
+
+    for (final prop in properties.where((p) => p.scope == DeclScope.public)) {
+      final spec =
+          prop.emit(options..override = isOverride(prop.dartName ?? prop.name));
+      if (spec is Method) {
+        methodDecs.add(spec);
+      } else {
+        fieldDecs.add(spec as Field);
+      }
+    }
+
+    methodDecs.addAll(methods.where((p) => p.scope == DeclScope.public).map(
+        (m) => m.emit(options!..override = isOverride(m.dartName ?? m.name))));
+    methodDecs.addAll(operators.where((p) => p.scope == DeclScope.public).map(
+        (m) => m.emit(options!..override = isOverride(m.dartName ?? m.name))));
+
+    final repType = this is ClassDeclaration
+        ? getClassRepresentationType(this as ClassDeclaration)
+        : BuiltinType.primitiveType(PrimitiveType.object, isNullable: false);
+
+    return ExtensionType((e) => e
+      ..name = dartName ?? name
+      ..annotations.addAll([
+        if (dartName != null && dartName != name) generateJSAnnotation(name)
+      ])
+      ..primaryConstructorName = '_'
+      ..representationDeclaration = RepresentationDeclaration((r) => r
+        ..declaredRepresentationType = repType.emit(options?.toTypeOptions())
+        ..name = '_')
+      ..implements.addAll([
+        if (extendees.isEmpty && implementees.isEmpty)
+          refer('JSObject', 'dart:js_interop')
+        else ...[
+          ...extendees.map((e) => e.emit(options?.toTypeOptions())),
+          ...implementees.map((i) => i.emit(options?.toTypeOptions()))
+        ]
+      ])
+      ..types
+          .addAll(typeParameters.map((t) => t.emit(options?.toTypeOptions())))
+      ..constructors.addAll([
+        if (!abstract)
+          if (constructors.isEmpty && this is ClassDeclaration)
+            ConstructorDeclaration.defaultFor(this).emit(options)
+          else
+            ...constructors.map((c) => c.emit(options))
+      ])
+      ..fields.addAll(fieldDecs)
+      ..methods.addAll(methodDecs));
+  }
 }
 
 abstract class MemberDeclaration {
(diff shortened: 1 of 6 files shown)


web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -72,6 +79,537 @@ class Transformer {
...
+
+    final outputInterface = InterfaceDeclaration(
+        name: name,
+        dartName: dartName,
+        id: id,
+        exported: isExported,
+        typeParameters:
+            typeParams?.map(_transformTypeParamDeclaration).toList() ?? [],
+        extendedTypes: extendees,
+        methods: List.empty(growable: true),
```
srujzs, line 121: Why not just `[]` here and elsewhere? Also, we can exclude these parameters from the constructor (and just have field initializers) if they're always going to be initialized to empty lists.

https://api.dart.dev/dart-core/List/List.empty.html

> If growable is true, the list is growable and equivalent to <E>[].

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -72,6 +79,537 @@ class Transformer {
     nodes.add(node);
   }
 
+  InterfaceDeclaration _transformInterface(TSInterfaceDeclaration interface) {
```
srujzs, line 82: This is also very similar to the code below in `_transformClass`, can we share the code?

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -484,32 +1000,80 @@ class Transformer {
...
-        });
+      case final CallableDeclaration f:
+        filteredDeclarations.addAll(getCallableDependencies(f));
         break;
       case final EnumDeclaration _:
         break;
       case final TypeAliasDeclaration t:
         if (decl.type is! BuiltinType) filteredDeclarations.add(t.type);
         break;
+      case final ClassDeclaration cl:
```
srujzs, line 1030: Note that this very similar to the below code. If we can move some of the shared fields up to `TypeDeclaration` like mentioned above, we can reduce the duplicate code.

web_generator/test/integration/interop_gen/classes_expected.dart
```diff
@@ -0,0 +1,322 @@
...
+@_i1.JS()
+external Sphere drawSphere(Point3D center);
+extension type EpahsImpl<TMeta extends _i1.JSAny?>._(_i1.JSObject _)
+    implements Epahs<TMeta> {
+  external EpahsImpl(
+    String name, [
+    AnonymousUnion$1? type,
+  ]);
+
+  external factory EpahsImpl.unnamed$1(Epahs<TMeta> config);
```
srujzs, line 281: Can this just be `$1` (as the unnamed is obvious)?

web_generator/test/integration/interop_gen/classes_input.d.ts
```diff
@@ -0,0 +1,177 @@
...
+    constructor(name: string, type?: 'circle' | 'rectangle' | 'polygon');
+    onUpdate?(prev: Epahs<TMeta>): void;
+    constructor(config: Epahs<TMeta>);
+    get location(): string;
+    set location(value: string);
+    area(): number;
+    area(unit: 'cm2' | 'in2'): string;
+    static getById(id: string): EpahsImpl;
+    toString(): string;
+}
```
srujzs, line 177: nit: newline

web_generator/lib/src/ast/base.dart
```diff
@@ -74,3 +76,39 @@ abstract class Type extends Node {
...
+enum DeclScope { private, protected, public }
+
+class ParameterDeclaration {
+  final String name;
+
+  final bool optional;
+
+  final Type type;
+
+  final bool variardic;
```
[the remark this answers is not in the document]
nikeokoronkwo: oops :(

web_generator/lib/src/ast/helpers.dart
```diff
@@ -50,3 +50,68 @@ List<Parameter> spreadParam(ParameterDeclaration p, int count) {
...
+
+Set<String> getMemberHierarchy(TypeDeclaration type) {
+  final members = <String>{};
+
+  switch (type) {
+    case ClassDeclaration(
+        extendedType: final extendee,
+        implementedTypes: final implementees
+      ):
+      if (extendee case ReferredType<Declaration>(declaration: final d)
```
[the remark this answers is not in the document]
nikeokoronkwo: That should be for the type hierarchy correct?
Could you elaborate a bit more, or maybe I am getting what you are referencing wrong

web_generator/lib/src/ast/declarations.dart
```diff
@@ -269,3 +268,539 @@ class TypeAliasDeclaration extends NamedDeclaration
...
+      this.dartName,
+      this.typeParameters = const [],
+      this.extendedTypes = const [],
+      this.methods = const [],
+      this.properties = const [],
+      this.operators = const [],
+      this.constructors = const []});
+
+  @override
+  ExtensionType emit([covariant DeclarationOptions? options]) {
```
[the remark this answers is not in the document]
nikeokoronkwo: Sorry, could you elaborate a bit more. Like you said, `code_builder` building once is the issue

web_generator/lib/src/ast/declarations.dart
```diff
@@ -269,3 +268,539 @@ class TypeAliasDeclaration extends NamedDeclaration
...
+      required this.static,
+      this.isNullable = false});
+
+  @override
+  Spec emit([covariant DeclarationOptions? options]) {
+    options ??= DeclarationOptions();
+    if (readonly) {
+      return Method((m) => m
+        ..external = true
+        ..name = '${scope == DeclScope.public ? '' : '_'}${dartName ?? name}'
```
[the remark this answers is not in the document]
nikeokoronkwo: It's just here for completeness.

web_generator/lib/src/ast/declarations.dart
```diff
@@ -9,14 +9,32 @@ import 'base.dart';
...
+
+  abstract final List<PropertyDeclaration> properties;
+
+  abstract final List<OperatorDeclaration> operators;
+}
+
+abstract class MemberDeclaration {
+  late final TypeDeclaration parent;
+
+  abstract final DeclScope scope;
```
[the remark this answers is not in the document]
nikeokoronkwo: You are right on this, but because of the way TS works, some declarations and members may be referenced in types for public members (indexed access members, `typeof`, etc), so skipping these types could cause edge case issues in those aspects

web_generator/test/integration/interop_gen/classes_expected.dart
nikeokoronkwo: This is because when we read the name of the constructor from TS it (for some reason) explicitly returns 'unnamed'

web_generator/lib/src/interop_gen/transform/transformer.dart
nikeokoronkwo: So have a case statement that cascades down?

web_generator/lib/src/ast/declarations.dart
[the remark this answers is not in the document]
srujzs: Maybe just assert that `scope` is `DeclScope.public` in `emit` so we don't assume it can flow through here (above the `if (readonly)` check) instead. That way we can just use `'${dartName ?? name}'` here. Same comment in `MethodDeclaration`.

web_generator/lib/src/ast/declarations.dart
```diff
@@ -269,3 +268,539 @@ class TypeAliasDeclaration extends NamedDeclaration
...
+      } else {
+        if (p.optional) {
+          optionalParams.add(p.emit(options));
+        } else {
+          requiredParams.add(p.emit(options));
+        }
+      }
+    }
+
+    if (isNullable) {
```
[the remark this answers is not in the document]
srujzs: Yeah, there are downsides to both approaches. Eventually, we'll have a generic type for `JSFunction` which should make this option less ugly.

So, I'm leaning towards keeping it as is for now (using a getter).

web_generator/lib/src/interop_gen/transform/transformer.dart
srujzs: To elaborate, the implementation of this method and `_transformClass` do very similar things and have a very similar structure, so it may help to have a helper method here.

web_generator/lib/src/interop_gen/transform/transformer.dart
srujzs: Talked offline, but if we had `case final ClassOrInterfaceDeclaration ci`, then we can have shared code. We could also cascade it like you suggested where we first check `case final TypeDeclaration` then `case final ClassOrInterfaceDeclaration`, etc.

The point is really just to have more shared code for code that handles `ClassDeclaration`s and `InterfaceDeclaration`s.

web_generator/test/integration/interop_gen/classes_expected.dart
srujzs: Something to check: maybe the `unnamed` is coming from the namer? If you think it's easiest to implement it this way, I don't mind too much.

srujzs (approved):

kevmoo:
good to merge @srujzs ?

srujzs (approved):
Looks like you were able to unify the various class and interface code, awesome!

PR #417, opened 2025-07-15
[interop] Support `typeof` type declarations

### LINKED ISSUE #383: Interop Gen: Support `typeof` type declarations
This issue is to add support for parsing types using `typeof` declarations. 

```ts
export const foo: Record<string, number> = {};
export const bar: typeof foo = {};
```

This adds support for a new `Type` declaration for handling such `typeof` declarations, and generating references to the given type when a reference is needed for such declaration.

nikeokoronkwo:
Fixes #383

This PR adds support for `typeof` types from TS into the Dart JS Interop Interface Generator, also supporting referencing the types from Functions, variables and typeable declarations.

Declarations with types referencing declarations inherit the type of the declaration, which may range from simply mapping the type name of the referenced type (for types like enums, classes, etc) to mapping the structure of the declaration for functions.

Much larger scope will be supported once qualified names are supported.

### DIFF
--- a/web_generator/lib/src/interop_gen/transform/transformer.dart
+++ b/web_generator/lib/src/interop_gen/transform/transformer.dart
@@ -51,6 +51,7 @@ class Transformer {
         filterDeclSet = filterDeclSet.toList(),
         namer = UniqueNamer();
 
+  /// Transforms a TypeScript AST Node [TSNode] into a Dart representable [Node]
   void transform(TSNode node) {
     if (nodes.contains(node)) return;
 
@@ -695,7 +696,16 @@ class Transformer {
         constraint: getJSTypeAlternative(constraint));
   }
 
-  /// Parses the type
+  /// Parses a TypeScript AST Type Node [TSTypeNode] into a [Type] Node
+  /// used to represent a type
+  ///
+  /// [parameter] represents whether the [TSTypeNode] is being passed in
+  /// the context of a parameter, which is mainly used to differentiate between
+  /// using [num] and [double] in the context of a [JSNumber]
+  ///
+  /// [typeArg] represents whether the [TSTypeNode] is being passed in the
+  /// context of a type argument, as Dart core types are not allowed in
+  /// type arguments
   Type _transformType(TSTypeNode type,
       {bool parameter = false, bool typeArg = false}) {
     switch (type.kind) {
@@ -799,7 +809,7 @@ class Transformer {
         final typeArguments = typeQuery.typeArguments?.toDart;
 
         return _getTypeFromDeclaration(exprName, typeArguments,
-            typeArg: typeArg, declarationOnly: true);
+            typeArg: typeArg, isNotTypableDeclaration: true);
       case TSSyntaxKind.ArrayType:
         return BuiltinType.primitiveType(PrimitiveType.array, typeParams: [
           getJSTypeAlternative(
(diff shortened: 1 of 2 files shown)


web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -419,6 +419,37 @@ class Transformer {
...
+          transform(declaration);
+
+          declarationsMatching = nodeMap.findByName(name);
+        }
+
+        final firstNode =
+            declarationsMatching.whereType<NamedDeclaration>().first;
+
+        return firstNode.asReferredType(
+          (typeArguments ?? [])
```
srujzs, line 449: nit: you can just apply the map to `typeArguments` e.g. `typeArguments.map(...).toList() ?? []`

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -419,6 +419,37 @@ class Transformer {
...
+        if (declarationsMatching.isEmpty) {
+          final symbol = typeChecker.getSymbolAtLocation(exprName);
+          final declarations = symbol?.getDeclarations();
+          final declaration = declarations?.toDart.first;
+
+          if (declaration == null) {
+            throw Exception('Found no declaration matching $name');
+          }
+
+          transform(declaration);
```
srujzs, line 440: Are there any potential pitfalls in doing this transform out-of-order? I assume not, but I suppose once we handle fully-qualified names in `typeof`, this might get a bit weird.

web_generator/test/integration/interop_gen/ts_typing_input.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    C = 2,
+    D = 4
+}
+export declare const myEnumValue: MyEnum;
+export declare const myEnumValue2: typeof MyEnum;
+export declare function myFunction(param: string): string;
+export declare let myFunctionAlias: typeof myFunction;
+export declare let myFunctionAlias2: typeof myFunctionAlias;
+export declare function myEnclosingFunction(func: typeof myFunction): string;
+export declare const myEnclosingFunctionAlias: typeof myEnclosingFunction;
```
srujzs, line 17: nit: newline

web_generator/test/integration/interop_gen/ts_typing_input.d.ts
```diff
@@ -0,0 +1,17 @@
+export declare const myString: string;
```
srujzs, line 1: Maybe have an example where we do a `typeof` of a declaration that is defined afterwards?

web_generator/test/integration/interop_gen/ts_typing_input.d.ts
```diff
@@ -0,0 +1,17 @@
...
+export declare const myNumberArray: number[];
+export declare const myCloneString: typeof myString;
+export declare const muCloneNumberArray: typeof myNumberArray[];
```
srujzs, line 4: nit: myCloneNumberArray

web_generator/lib/src/interop_gen/transform/transformer.dart
nikeokoronkwo: Could you elaborate what you mean by out-of-order? The transformation is intended to be done once it is fetched so that it is guaranteed that the declaration exists before re-fetching.

For fully qualified names, we will need to fetch the declaration and probably do recursive lookup from there depending on what declaration it is

srujzs: > The transformation is intended to be done once it is fetched so that it is guaranteed that the declaration exists before re-fetching.

Ah okay, this is what I was wondering about. SGTM to worry about the fully qualified names in a separate PR.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -840,7 +846,7 @@ class Transformer {
 
   Type _getTypeFromDeclaration(
       TSIdentifier typeName, List<TSTypeNode>? typeArguments,
-      {bool typeArg = false}) {
+      {bool typeArg = false, bool declarationOnly = false}) {
```
srujzs, line 849: What is the purpose of this parameter? Can we have a better name/documentation?

srujzs (commented):
Thanks! Just one final comment about the `declarationOnly` addition.

srujzs (approved):
Thanks for the documentation, LGTM!

PR #418, opened 2025-07-17
[interop] Add support for importing and exporting declarations, as well as multi-file output

### LINKED ISSUE #378: Interop Gen: Support Export/Import Declarations
This issue is to add support to the Dart JS Interop Generator for supporting imported/exported declarations.

Some declaration files may export declarations from other files, and we should handle these declarations, most likely by parsing the necessary file, filtering for the specified declaration, and associating any aliases placed on the declaration.

This issue also addresses handling default exports, defaulting their names to `default` when not named.

```ts
// a.ts
export { a, b, c } // named items
export default c; // named default export
export * from "file.ts" // export from file

// b.ts
import { a as z } from "a.ts";

export { z } // aliasing
```

### LINKED ISSUE #387: Interop Gen: Support Import Declarations and Multi-File Output
This issue adds support for generating code for imported statements in a declaration file.

If the code is for a single-output file, then only the declaration and its dependencies are generated, then returned. If multi-output, the whole file is imported, and then used as a referenced imported declaration.

nikeokoronkwo:
Fixes #378 
Fixes #387 

This PR is for adding support for multiple file processing, with support for exporting and importing declarations even between files. 

It makes use of a `TransformManager` which transforms various files and uses a `ProgramMap` to memoize and keep track of dependencies across files, which helps in reducing expensive transform calls to declarations already referenced from other files.
Rather than transforming every declaration in a source file, the type checker gets the declarations exported from the file, and processes those declarations and its dependencies only, making processing faster

The type checker helps to resolve types and imports, as well as resolve import declarations. Import declarations are not parsed by default so that we know we are only making use of imports actually referenced in code (which is helpful in single outputs for instance).

(The program map makes use of `package:path`s [`PathMap`](https://pub.dev/documentation/path/latest/path/PathMap-class.html) to easily work with paths for files and declarations, as the TS output may resolve either to a relative import in the file, or may reference the declaration externally as an absolute path).

Imported declarations are declared as `ReferredType`s with a `url` property, which allows for referencing the normal

Exports are rewritten based on how they are specified, and are processed after transformation (since they are separate declarations and may be referenced as they are.

Imports are processed depending on the presence of the file:
- If TS can find the file (i.e not global scope), then the declaration is referenced with the proper name as an import
- If the file needed/imported is not included in the multi-file input params, then the declarations are generated in the output file as a single output.
- If the declaration is in the global scope, then the generator can auto-generate these declarations as so.


web_generator/lib/src/js/helpers.dart
```diff
@@ -0,0 +1,140 @@
+import 'dart:collection';
+import 'dart:js_interop';
+
```
srujzs, line 3: Let's leave a comment/TODO with the bug you filed to add these types to `dart:js_interop` to remove these once that's done.

(super cool to see some of this implementation btw)

web_generator/test/integration/interop_gen/import_export_input.d.ts
```diff
@@ -0,0 +1,49 @@
...
+export declare const unitK3D: Vector3D;
+export {
+    origin, origin3D, dotProduct, crossProduct, mapTo3D,
+    TransformerMatrix, ComparatorMatrix
+}
+export {
+    PolarCoordinate as PolarPoint, 
+    CylindricalCoordinate as CylindricalPoint,
+    SphericalCoordinate as SphericalPoint,
+    toPolarCoordinate,
```
srujzs, line 46: Maybe rename a member in the export to see what it might look like?

web_generator/test/integration/interop_gen/project/input/e.d.ts
```diff
@@ -0,0 +1,27 @@
+export declare class Configuration {
```
srujzs, line 1: I'm guessing the purpose of this file is just so it can be correctly ignored?

What happens if a file imports this file and exports some of its members but it's not in the config? Is there an error in the generator?

web_generator/lib/src/ast/declarations.dart
```diff
@@ -160,17 +160,20 @@ class VariableDeclaration extends FieldDeclaration
 
   @override
```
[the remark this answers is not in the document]
srujzs: Hmm, can we have an abstract "set dartName" in `ExportableDeclaration` then? As only those declarations can possibly be renamed?

web_generator/lib/src/interop_gen/transform.dart
```diff
@@ -77,45 +82,154 @@ extension type NodeMap._(Map<String, Node> decls) implements Map<String, Node> {
...
+    // check
+    NodeMap nodeMap;
+    if (_map.containsKey(file)) {
+      nodeMap = _map[file]!;
+    } else {
+      final src = program.getSourceFile(file);
+
+      // TODO: We could add the exported declarations here ahead of
+      //  time thanks to the type checker.
+      //  @srujzs What do you think?
```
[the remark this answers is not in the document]
nikeokoronkwo: What we can do instead is to perform recursive transformation on the declarations exported, which would avoid us having to parse declarations that won't be exported, nor appear as dependencies, and then return. 

Ideally this should reduce the amount of calls being made (improving overall performance), as it would scope filtering and dependency filtering (filtering would only occur based on user-specified names, so if `include` and `exclude` are empty, then there should be no double pass).

web_generator/test/integration/interop_gen/import_export_input.d.ts
srujzs: Sorry, I meant rename a member as in a method. We only rename classes here I believe.

web_generator/test/integration/interop_gen/project/input/e.d.ts
srujzs: Ooh interesting. That sounds good to me then. An error would prevent confusion but I think it's fine to somewhat handle this.

web_generator/lib/src/interop_gen/transform.dart
```diff
@@ -74,48 +79,172 @@ extension type NodeMap._(Map<String, Node> decls) implements Map<String, Node> {
...
+  /// The type checker for the given program
+  ///
+  /// It is generated as this to prevent having to regenerate it multiple times
+  final ts.TSTypeChecker typeChecker;
+
+  /// The files in the given project
+  final List<String> files;
+
+  List<String> get absoluteFiles =>
+      files.map((f) => p.normalize(p.absolute(f))).toList();
```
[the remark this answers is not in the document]
nikeokoronkwo: Do you mean just to have one of the fields (i.e just `absoluteFiles`)?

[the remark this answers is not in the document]
srujzs: Sure! If `files` is unused, then we could just have `absoluteFiles` as a singular field.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -984,58 +1046,81 @@ class Transformer {
...
+    }
+
+    return asReferredType;
   }
 
-  Type? _deriveTypeOrTransform(TSSymbol symbol,
-      [String? name, List<TSTypeNode>? typeArguments, bool typeArg = false]) {
+  /// Either derives the type referenced by the [symbol], or
+  /// transforms the declaration(s) associated with the [symbol].
+  ///
```
srujzs, line 1104: nit: stray comment markers, perhaps you meant to document the return value?

srujzs (approved):

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -115,6 +123,23 @@ class Transformer {
...
+
+    final dartName = specifier.name;
+
+    final decl = nodeMap.findByName(actualName.text);
+
+    // This just guarantees the declaration is transformed before adding the
+    // export reference
+    if (decl.isEmpty) _getTypeFromDeclaration(actualName, []);
+
+    exportSet.removeWhere((e) => e.name == actualName.text);
```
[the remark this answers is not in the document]
nikeokoronkwo: Export specifiers are contained in export declarations. There can be multiple export declarations. The order shouldn't matter, as you cannot export a symbol twice.

PR #423, opened 2025-07-20
[web_generator] Bug Fixes in Entrypoint `gen_interop_bindings.dart`

nikeokoronkwo:
- removed duplicate `help` option
- added support for no-config when running


srujzs (approved):
I wasn't added for review, but I'm guessing we're good to merge?

nikeokoronkwo:
> Is that a recent change? Or have I just not been noticing? :P

It's been like that since I started. I have to wait and hope you come across it :)

PR #426, opened 2025-07-21
[interop] Implement Diagnostics and Handling Errors

### LINKED ISSUE #421: Interop Gen: Implement Error Diagnostics when parsing input files
This issue is to implement compiler diagnostics from `tsc` when parsing input files to prevent us from transforming/generating code for erroneous TypeScipt Files

nikeokoronkwo:
Fixes #421 

This PR adds support for handling diagnostics and errors when parsing TS declaration files for the generator to transform, preventing the generator from handling invalid TS files. 

Diagnostics are printed in the following format

```log
<filename> (<line no>,<char no>): <message>
```

And the process exits with status code 1.

### DIFF
--- a/web_generator/test/assets/invalid.d.ts
+++ b/web_generator/test/assets/invalid.d.ts
@@ -1,0 +1,5 @@
+// invalid-syntax.d.ts
+declare interface Person {
+  name: string
+  age number  // ❌ missing colon
+}
--- a/web_generator/test/assets/invalid.d.ts
+++ b/web_generator/test/assets/invalid.d.ts
@@ -1,5 +1,12 @@
 // invalid-syntax.d.ts
-declare interface Person {
+export interface Person {
   name: string
-  age number  // ❌ missing colon
+  age number
+}
+interface User {
+  id: string;
+}
+export declare class Admin implements User {
+  constructor(name: string);
+  toString(): string;
 }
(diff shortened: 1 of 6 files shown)


web_generator/test/assets/invalid.d.ts
```diff
@@ -0,0 +1,5 @@
+// invalid-syntax.d.ts
+declare interface Person {
+  name: string
+  age number  // ❌ missing colon
```
srujzs, line 4: Should we test semantic and declaration diagnostics as well?

srujzs (approved):
Nice, thanks!

web_generator/test/invalid_input_test.dart
```diff
@@ -0,0 +1,49 @@
...
+            'main.mjs',
+            '--input=$inputFilePath',
+            '--output=$outputFilePath',
+            '--declaration'
+          ],
+          workingDirectory: bindingsGenPath);
+
+      final stderr = await process.stderr.transform(utf8.decoder).toList();
+
+      expect(stderr, isNotEmpty);
```
srujzs, line 45: nit: Should we check the output or would that make this test too volatile?

srujzs: Ack, LGTM then. I'll land this.

PR #427, opened 2025-07-22
[web_generator] Add support for passing files as globs

nikeokoronkwo:
This PR adds support to passing files as globs for both the IDL generator and the TS Declarations generator via Node's [`fs.globSync`](https://nodejs.org/api/fs.html#fsglobsyncpattern-options) function.

### DIFF
--- a/web_generator/lib/src/config.dart
+++ b/web_generator/lib/src/config.dart
@@ -2,10 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:js_interop';
+
 import 'package:dart_style/dart_style.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart';
+
+import 'js/filesystem_api.dart';
 
 class FunctionConfig {
   /// The number of variable arguments
@@ -154,10 +158,13 @@ class YamlConfig implements Config {
       throw TypeError();
     }
 
+    final allFiles = fs.globSync(inputFiles.map((i) => i.toJS).toList().toJS,
+        FSGlobSyncOptions(cwd: p.dirname(filename).toJS));
+
     return YamlConfig._(
         filename: Uri.file(filename),
-        input: inputFiles
-            .map((file) => p.join(p.dirname(filename), file))
+        input: allFiles.toDart
+            .map((file) => p.join(p.dirname(filename), file.toDart))
             .toList(),
         output:
             p.join(p.dirname(filename), (yaml['output'] ?? output) as String),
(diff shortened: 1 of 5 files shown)


web_generator/lib/src/config.dart
```diff
@@ -174,3 +185,14 @@ class YamlConfig implements Config {
             []);
   }
 }
+
+/// Creates the `exclude` function for [FSGlobSyncOptions]'s `exclude` option
+/// to exclude all files not ending with the given extension
```
srujzs, line 190: nit: `[extension]`

web_generator/lib/src/config.dart
```diff
@@ -174,3 +185,14 @@ class YamlConfig implements Config {
...
+///
+/// This helps support passing dir globs.
+bool Function(FSDirent entry) excludeFileEntryFunc(String extension) {
```
srujzs, line 193: nit: `allowedExtension` instead of `extension` so it's obvious that it's being included not excluded

srujzs (commented):
Feel free to update the node version we use if you think that may be the issue (I believe that shouldn't result in any issues for `package:web`)

nikeokoronkwo:
> Feel free to update the node version we use if you think that may be the issue

Does `mono_repo` support set up actions? That way I can update it globally

devoncarew:
> @devoncarew Is there a better way to specify a newer Node version that I'm missing?

I've largely moved away from using mono_repo; it adds a layer between the user and the actual CI definition that adds additional complexity, makes simple things harder, and doesn't really have additional value over just raw github actions.

I think it would be worth just adding a new workflow definitions per package in this repo (2?) and removing the mono_repo config info. Then things like specifying node versions and such should be straightforward. I can point you to some same workflows if helpful.

srujzs:
> I've largely moved away from using mono_repo; it adds a layer between the user and the actual CI definition that adds additional complexity, makes simple things harder, and doesn't really have additional value over just raw github actions.

> I think it would be worth just adding a new workflow definitions per package in this repo (2?) and removing the mono_repo config info. Then things like specifying node versions and such should be straightforward. I can point you to some same workflows if helpful.

That makes sense. Do you have an example where you migrated away from `mono_repo.yml`? I'm tempted to keep `dart.yml` as the source of truth but it's quite messy and I wonder if there's cruft that can be removed.

For now, I believe @nikeokoronkwo will just `glob` things early so we don't need a new node version, but I'd like to move away from `mono_repo.yml` at some point as well.

nikeokoronkwo:
Yeah, I can't quite get to why the builds are failing: the tests work on my system and they've been working up until now.

web_generator/lib/src/config.dart
```diff
@@ -154,10 +158,17 @@ class YamlConfig implements Config {
       throw TypeError();
     }
 
+    final allFiles = fs.globSync(
```
srujzs, line 161: nit: The `globSync` code seems to be fairly similar everytime `globSync` is used, maybe move it to a helper function?

PR #428, opened 2025-07-23
switch CI to using workflow files

devoncarew:
- switch CI to using workflow files

---

<details>

</details>


.github/workflows/web_generator.yaml
```diff
@@ -0,0 +1,67 @@
...
+  run:
+    working-directory: web_generator/
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sdk: [stable]
```
srujzs, line 28: We were testing `dev` before so maybe replace this with that? We could include `stable` if we want, but `dev` is more important since that's what users (us!) will likely use for this package.

.github/workflows/web.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
+        with:
+          sdk: ${{ matrix.sdk }}
+
+      - run: dart pub get
+      - run: dart format --output=none --set-exit-if-changed .
+        if: ${{ matrix.sdk == 'stable' }}
```
srujzs, line 39: We ran `dart format` with dev as well before. I'd remove this (and it's okay if it runs with beta).

.github/workflows/web_generator.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
+        with:
+          sdk: ${{ matrix.sdk }}
+
+      - run: dart pub get
+      - run: dart format --output=none --set-exit-if-changed .
+        if: ${{ matrix.sdk == 'stable' }}
```
srujzs, line 39: We were formatting with `dev` before so I'd just remove.

devoncarew (commented):
Thanks for the review! Updated based on the feedback. I did keep the sdk version check for the formatting verification, just so we're only checking for formatting against a single sdk version.

.github/workflows/web_generator.yaml
devoncarew: (actually, I removed - as you said, we're only running this workflow on `dev`)

srujzs (approved):
LGTM, thanks again!

Issue #380, opened 2025-06-10
Add support and create Dart representation for primitive JS types not currently supported in Dart


nikeokoronkwo:
This issue is for adding support for generating certain declarations for types found in the global scope for JS code, but not supported directly in `dart:js_interop` at the moment. These declarations may also be followed by helpers to help convert them to similar Dart types where possible. 

Some of these types are native to JS and some are supported by TS. For those supported by JS natively, we will need to create respective `external` calls, while those supported by TS will need to have representations that match the compilation result in JS:

- [ ] `Iterable` and `Iterator`: See [dart-lang/sdk#60457](https://github.com/dart-lang/sdk/issues/60457)
- [ ] `Generator`: This can be converted to/from an `Iterable` 
- [ ] `AsyncGenerator`: This can be converted to/from a `Stream` and be used with `await for`
- [ ] `Record`: TS Native, also mapped with Mapped Type
- [ ] `Map`
- [ ] `Set`
- [ ] The `Object` singleton
- [ ] `Error`: There are not may instances of this standalone
- [x] `URL`: This can be converted to/from a Dart `Uri`
- [ ] Regex: We should support literal regexes, and be able to convert to/from dart Regex
- [x] #433
- [ ] `WeakMap`
- [ ] `WeakSet`

PR #434, opened 2025-08-02
[interop] Add Support for Anonymous Declarations

### LINKED ISSUE #385: Interop Gen: Support Union Types
This issue is to add support for parsing and transforming union types in typescript for the Dart JS Interop Generator

```ts
export type Color = 'red' | 'green' | 'blue';
export declare const stringOrNumber: string | number;

export declare const colorOrOther: Color | string;
```

Some union types can be reduced to base interop types such as `JSString` for the typical "enum"-like union syntax (`'red' | 'green' | 'blue'`), `JSObject`, or `JSAny`.

## Custom Types

We are considering generating custom types containing an intersection of the properties contained in the union of the types (if any). We can make use of the typescript type checker API to find an intersection for these types, and add helpers for casting to the necessary union type, as well as checking if such type is one or the other. 

We would also need to consider handling `isA` checks if we are going on with custom types, as they do not exist on JS.

This option should be configurable as well, as users may choose to generate such custom type or not (i.e default to base intersection type)

### LINKED ISSUE #410: Interop Gen: Support Constructor Types
This issue is to add support for TypeScript Constructor Types in the Dart TS Interop Interface Generator.

```ts
// assuming there is a class called `MyClass` defined
export declare const myConstructor: new () => MyClass;
```

This will be helpful when working with namespaces (see [#399](https://github.com/dart-lang/web/issues/399)), as translating classes defined in namespaces would mean interfacing class calls as constructor type calls.

### LINKED ISSUE #422: Interop Gen: Support Function Types

### LINKED ISSUE #433: Tuples: See #385

nikeokoronkwo:
Fixes #385 
Fixes #422 
Fixes #410 
Fixes #433

This PR adds support for generating declarations for anonymous declarations, such as anonymous objects, 

A hashing function is used for consistently hashing string objects, which is used for hashing identifiers for anonymous unions, objects and more, as well as comparing such names to allow reusing of such types.

- Anonymous Unions are turned into a Dart (anonymous) extension type which has a rep type/implements the shared type between the union parts. It has getters to get the value as a certain probability of the union. For instance, a union `string | number | boolean` may look like the following:
    ```dart
    extension type _AnonymousUnion_204398._(JSAny _) implements JSAny {
      String get asString => (_ as JSString).toDart;
      double get asDouble => (_ as JSNumber).toDartDouble;
      bool get asBool => (_ as JSBoolean).toDart;
    }
    ```
- Anonymous Tuples are turned into an instance of a `JSTupleX` declaration generated depending on the number of members it may have (i.e if a 2 member tuple exists, it will be an instance of `JSTuple2<A, B>`). The tuple extension type looks like the following:
    ```dart
    // normal
    extension type JSTupleX<A, B, ..., B1>._(JSArray<JSAny> _) implements JSArray<JSAny> {
      A get $1 => _[0] as A;
      B get $2 => _[1] as B;
      /* ... */
      B1 get $x => _[x] as B1;
      
      set $1(A newValue) => _[0] = newValue;
      set $2(B newValue) => _[1] = newValue;
      /* ... */
      set $x(B1 newValue) => _[x] = newValue;
    }
    
    // readonly
    extension type JSReadonlyTupleX<A, B, ..., B1>._(JSArray<JSAny> _) implements JSArray<JSAny> {
      A get $1 => _[0] as A;
      B get $2 => _[1] as B;
      /* ... */
      B1 get $x => _[x] as B1;
    }
    ```

## TODO
- [ ] Support Subtype Deduction
- [x] Support Anonymous Unions
- [ ] Support Anonymous Objects
- [x] Support Anonymous Tuples
- [ ] Support Anonymous Closures
- [ ] Support Anonymous Constructor Signatures

### DIFF
--- a/web_generator/lib/src/ast/base.dart
+++ b/web_generator/lib/src/ast/base.dart
@@ -86,6 +86,14 @@ abstract class Type extends Node {
 
   @override
   Reference emit([covariant TypeOptions? options]);
+
+  @override
+  bool operator ==(Object other) {
+    return other is Type && id == other.id;
+  }
+
+  @override
+  int get hashCode => Object.hashAll([id]);
 }
 
 abstract class FieldDeclaration extends NamedDeclaration {
@@ -101,6 +109,12 @@ abstract class CallableDeclaration extends NamedDeclaration {
 }
 
 enum DeclScope { private, protected, public }
+
+abstract class DeclarationAssociatedType<T extends Declaration> extends Type {
+  T get declaration;
+
+  String get declarationName;
+}
 
 class ParameterDeclaration {
   final String name;
--- a/web_generator/lib/src/ast/builtin.dart
+++ b/web_generator/lib/src/ast/builtin.dart
@@ -81,7 +81,10 @@ class BuiltinType extends Type {
               name: 'JSString', fromDartJSInterop: true, isNullable: isNullable)
           : BuiltinType(name: 'String', isNullable: isNullable),
       PrimitiveType.$void || PrimitiveType.undefined => $voidType,
-      PrimitiveType.any || PrimitiveType.unknown => anyType,
+      PrimitiveType.any => (isNullable ?? false)
+          ? anyType
+          : BuiltinType(name: 'JSAny', fromDartJSInterop: true),
+      PrimitiveType.unknown => anyType,
       PrimitiveType.object => BuiltinType(
           name: 'JSObject', fromDartJSInterop: true, isNullable: isNullable),
       PrimitiveType.symbol => BuiltinType(
--- a/web_generator/lib/src/ast/declarations.dart
+++ b/web_generator/lib/src/ast/declarations.dart
@@ -45,11 +45,14 @@ sealed class TypeDeclaration extends NamedDeclaration
       this.operators = const [],
       this.constructors = const []});
 
+  /// [assertRepType] is used to assert that the extension type generated
+  /// has a representation type of the first member of [extendees] if any.
   ExtensionType _emit(
       [covariant DeclarationOptions? options,
       bool abstract = false,
       List<Type> extendees = const [],
-      List<Type> implementees = const []]) {
+      List<Type> implementees = const [],
+      bool assertRepType = false]) {
     options ??= DeclarationOptions();
 
     final hierarchy = getMemberHierarchy(this);
@@ -75,8 +78,8 @@ sealed class TypeDeclaration extends NamedDeclaration
     methodDecs.addAll(operators.where((p) => p.scope == DeclScope.public).map(
         (m) => m.emit(options!..override = isOverride(m.dartName ?? m.name))));
 
-    final repType = this is ClassDeclaration
-        ? getClassRepresentationType(this as ClassDeclaration)
+    final repType = assertRepType || this is ClassDeclaration
+        ? getClassRepresentationType(this)
         : BuiltinType.primitiveType(PrimitiveType.object, isNullable: false);
 
     return ExtensionType((e) => e
@@ -114,6 +117,8 @@ abstract class MemberDeclaration {
   late final TypeDeclaration parent;
 
   abstract final DeclScope scope;
+
+  String? get name;
 }
 
 class VariableDeclaration extends FieldDeclaration
@@ -423,6 +428,11 @@ class InterfaceDeclaration extends TypeDeclaration {
 
   final List<Type> extendedTypes;
 
+  /// This asserts that the extension type generated produces a rep type
+  /// other than its default, which is denoted by the first member of
+  /// [extendedTypes] if any.
+  final bool assertRepType;
+
   InterfaceDeclaration(
       {required super.name,
       required super.exported,
@@ -433,15 +443,12 @@ class InterfaceDeclaration extends TypeDeclaration {
       super.methods,
       super.properties,
       super.operators,
-      super.constructors});
+      super.constructors,
+      this.assertRepType = false});
 
   @override
   ExtensionType emit([covariant DeclarationOptions? options]) {
-    return super._emit(
-      options,
-      false,
-      extendedTypes,
-    );
+    return super._emit(options, false, extendedTypes, [], assertRepType);
   }
 }
 
--- a/web_generator/lib/src/ast/helpers.dart
+++ b/web_generator/lib/src/ast/helpers.dart
@@ -4,6 +4,7 @@
 
 import 'package:code_builder/code_builder.dart';
 
+import '../interop_gen/namer.dart';
 import 'base.dart';
 import 'builtin.dart';
 import 'declarations.dart';
@@ -96,11 +97,19 @@ Set<String> getMemberHierarchy(TypeDeclaration type,
   return members;
 }
 
-Type getClassRepresentationType(ClassDeclaration cl) {
-  if (cl.extendedType case final extendee?) {
+Type getClassRepresentationType(TypeDeclaration cl) {
+  if (cl case ClassDeclaration(extendedType: final extendee?)) {
     return switch (extendee) {
-      final ClassDeclaration classExtendee =>
-        getClassRepresentationType(classExtendee),
+      ReferredType(declaration: final d) when d is ClassDeclaration =>
+        getClassRepresentationType(d),
+      final BuiltinType b => b,
+      _ => BuiltinType.primitiveType(PrimitiveType.object, isNullable: false)
+    };
+  } else if (cl case InterfaceDeclaration(extendedTypes: [final extendee])) {
+    return switch (extendee) {
+      ReferredType(declaration: final d) when d is ClassDeclaration =>
+        getClassRepresentationType(d),
+      final BuiltinType b => b,
       _ => BuiltinType.primitiveType(PrimitiveType.object, isNullable: false)
     };
   } else {
@@ -111,4 +120,32 @@ Type getClassRepresentationType(ClassDeclaration cl) {
 
     return BuiltinType.primitiveType(primitiveType, isNullable: false);
   }
+}
+
+Declaration generateTupleDeclaration(int count, {bool readonly = false}) {
+  final name = readonly ? 'JSReadonlyTuple$count' : 'JSTuple$count';
+  return InterfaceDeclaration(
+      name: name,
+      exported: true,
+      assertRepType: true,
+      id: ID(type: 'tuple', name: name),
+      typeParameters: List.generate(
+          count,
+          (index) => GenericType(
+              name: String.fromCharCode(65 + index),
+              constraint: BuiltinType.anyType)),
+      extendedTypes: [
+        BuiltinType.primitiveType(PrimitiveType.array,
+            shouldEmitJsType: true,
+            // TODO: get sub type instead
+            typeParams: [BuiltinType.anyType])
+      ],
+      properties: List.generate(
+          count,
+          (index) => PropertyDeclaration(
+              name: '\$${index + 1}',
+              id: ID(type: 'var', name: '\$${index + 1}'),
+              type: GenericType(name: String.fromCharCode(65 + index)),
+              static: false,
+              readonly: readonly)));
 }
--- a/web_generator/lib/src/ast/types.dart
+++ b/web_generator/lib/src/ast/types.dart
@@ -4,6 +4,7 @@
 
 import 'package:code_builder/code_builder.dart';
 import '../interop_gen/namer.dart';
+import '../interop_gen/sub_type.dart';
 import 'base.dart';
 import 'builtin.dart';
 import 'declarations.dart';
@@ -60,11 +61,46 @@ class ReferredDeclarationType<T extends Declaration> extends ReferredType<T> {
   }
 }
 
-// TODO(https://github.com/dart-lang/web/issues/385): Implement Support for UnionType (including implementing `emit`)
-class UnionType extends Type {
+class TupleType extends Type {
   final List<Type> types;
 
-  UnionType({required this.types});
+  TupleType({required this.types});
+
+  @override
+  ID get id => ID(type: 'type', name: types.map((t) => t.id.name).join(','));
+
+  @override
+  String? get name => null;
+
+  @override
+  Reference emit([TypeOptions? options]) {
+    return TypeReference((t) => t
+      ..symbol = 'JSTuple${types.length}'
+      ..types.addAll(types.map((type) => type.emit(options))));
+  }
+
+  @override
+  int get hashCode => Object.hashAllUnordered(types);
+
+  @override
+  bool operator ==(Object other) {
+    return other is TupleType && other.types.every(types.contains);
+  }
+}
+
+class UnionType extends Type implements DeclarationAssociatedType {
+  final List<Type> types;
+
+  final bool isNullable;
+
+  @override
+  String declarationName;
+
+  UnionType({
+    required this.types, 
+    required String name, 
+    this.isNullable = false
+  }) : declarationName = name;
 
   @override
   ID get id => ID(type: 'type', name: types.map((t) => t.id.name).join('|'));
@@ -73,16 +109,27 @@ class UnionType extends Type {
   String? get name => null;
 
   @override
+  Declaration get declaration =>
+      _UnionDeclaration(name: declarationName, types: types);
+
+  @override
   Reference emit([TypeOptions? options]) {
-    throw UnimplementedError('TODO: Implement UnionType.emit');
+    return TypeReference((t) => t
+      ..symbol = declarationName
+      ..isNullable = options?.nullable ?? isNullable);
+  }
+
+  @override
+  int get hashCode => Object.hashAllUnordered(types);
+
+  @override
+  bool operator ==(Object other) {
+    return other is TupleType && other.types.every(types.contains);
   }
 }
 
-// TODO: Handle naming anonymous declarations
-// TODO: Extract having a declaration associated with a type to its own type
-//  (e.g DeclarationAssociatedType)
 class HomogenousEnumType<T extends LiteralType, D extends Declaration>
-    extends UnionType {
+    extends UnionType implements DeclarationAssociatedType {
   final List<T> _types;
 
   @override
@@ -90,17 +137,14 @@ class HomogenousEnumType<T extends LiteralType, D extends Declaration>
 
   final Type baseType;
 
-  final bool isNullable;
-
-  String declarationName;
-
   HomogenousEnumType(
-      {required List<T> types, this.isNullable = false, required String name})
-      : declarationName = name,
-        _types = types,
-        baseType = types.first.baseType,
-        super(types: types);
+      {required List<T> super.types,
+      super.isNullable,
+      required super.name})
+      : _types = types,
+        baseType = types.first.baseType;
 
+  @override
   EnumDeclaration get declaration => EnumDeclaration(
       name: declarationName,
       dartName: UniqueNamer.makeNonConflicting(declarationName),
@@ -115,13 +159,6 @@ class HomogenousEnumType<T extends LiteralType, D extends Declaration>
         );
       }).toList(),
       exported: true);
-
-  @override
-  Reference emit([TypeOptions? options]) {
-    return TypeReference((t) => t
-      ..symbol = declarationName
-      ..isNullable = options?.nullable ?? isNullable);
-  }
 }
 
 /// The base class for a type generic (like 'T')
@@ -144,6 +181,16 @@ class GenericType extends Type {
     ..symbol = name
     ..bound = constraint?.emit()
     ..isNullable = options?.nullable);
+
+  @override
+  bool operator ==(Object other) {
+    return other is GenericType &&
+        other.name == name &&
+        other.constraint == constraint;
+  }
+
+  @override
+  int get hashCode => Object.hash(name, constraint);
 }
 
 /// A type representing a bare literal, such as `null`, a string or number
@@ -176,6 +223,14 @@ class LiteralType extends Type {
 
   @override
   ID get id => ID(type: 'type', name: name);
+
+  @override
+  bool operator ==(Object other) {
+    return other is LiteralType && other.name == name && other.value == value;
+  }
+
+  @override
+  int get hashCode => Object.hash(name, value);
 }
 
 enum LiteralKind {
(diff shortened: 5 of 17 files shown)


web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -1072,34 +1229,75 @@ class Transformer {
...
+          final homogenousEnumType =
+              HomogenousEnumType(types: nonNullLiteralTypes, name: name);
+
+          final homogenousEnum = typeMap.putIfAbsent(expectedId.toString(), () {
+            namer.markUsed(name);
+            return homogenousEnumType;
+          }) as HomogenousEnumType;
+
+          return homogenousEnum..isNullable = shouldBeNullable;
+        } else {
```
srujzs, line 1256: There's a lot of duplication in the `else` branch. Can we move the shared code outside of the branch instead?

web_generator/test/hasher_test.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:test/test.dart';
+
+void main() {
```
srujzs, line 3: There's no contents. :) Did you mean to add tests here?

web_generator/test/integration/interop_gen/namespaces_expected.dart
```diff
@@ -292,7 +293,7 @@ extension type EnterpriseApp_DataServices_ProductService._(_i1.JSObject _)
   external void save(EnterpriseApp_Models_Product item);
   external void add(EnterpriseApp_Models_Product product);
   @_i1.JS('get')
-  external EnterpriseApp_Models_Product get$(num id);
+  external EnterpriseApp_Models_Product get$(_AnonymousUnion_1467782 id);
```
srujzs, line 296: Hmm, I guess the parameter case is interesting here. I'm guessing users will want to create such a union when passing the value, so maybe we shouldn't make this private in the parameter case.

We may also need to add static `from_` methods e.g. `AnonymousUnion_2141241.fromString(String val) => AnonymousUnion_2141241._(val.toJS);` for these types.

web_generator/test/integration/interop_gen/ts_typing_expected.dart
```diff
@@ -40,7 +47,231 @@ external _i1.JSFunction get myEnclosingFunctionAlias;
...
+external _AnonymousUnion_5411652 get myUnion;
+@_i1.JS()
+external _AnonymousUnion_5411652 get myCloneUnion;
+@_i1.JS()
+external _AnonymousUnion_3781934 get mySecondUnion;
+@_i1.JS()
+external bool? get booleanOrUndefined;
+@_i1.JS()
+external _AnonymousUnion_6216705? get image;
+extension type JSTuple2<A extends _i1.JSAny?, B extends _i1.JSAny?>._(
```
srujzs, line 60: Do we make sure that we don't regenerate tuple types if they are referenced in another file? Should we instead generate these types in a separate file by themselves?

A better option might be to have a file that we always include that contains N such tuples so we don't have to always codegen them. We may at some point decide to expose that as part of other types like `JSMap`, `JSSet`, etc.

web_generator/test/integration/interop_gen/ts_typing_expected.dart
```diff
@@ -40,7 +47,231 @@ external _i1.JSFunction get myEnclosingFunctionAlias;
...
+  external double id;
+
+  external String name;
+
+  external double price;
+}
+@_i1.JS()
+external _AnonymousUnion_1189263 get responseObject;
+@_i1.JS()
+external _AnonymousConstructor_1059824 get productConstr;
```
srujzs, line 120: Oh I may have miscommunicated earlier what I meant, but I was thinking of making these constructor values like 

```dart
Product productConstr(
    num id,
    String name,
    num price,
  ) =>
      Product(
        id: id.toDouble(),
        name: name,
        price: price.toDouble(),
      );
```

instead. I'm not really sure it's useful for users to have this constructor as a getter (and it avoids the extra anonymous extension type if we don't), but I'm also totally okay with going the current route if you feel like it's better.

web_generator/test/integration/interop_gen/ts_typing_input.d.ts
```diff
@@ -17,9 +17,59 @@ export declare const myEnumValue2: typeof MyEnum;
...
+export declare const applyDiscount: (originalPrice: number) => number;
+export declare const shoppingCart: {
+    calculatedPrice: number;
+    displayInfo: () => void;
+    id: number;
+    name: string;
+    price: number;
+}[];
+export declare const createLogger: (prefix: string) => (message: string) => void;
+export declare const appLogger: (message: string) => void;
```
srujzs, line 75: nit: newline

web_generator/test/type_map_test.dart
```diff
@@ -0,0 +1,131 @@
...
+
+      final cfType =
+          getSubTypeOfTypes([c.asReferredType(), f.asReferredType()]);
+      expect(cfType, isA<ReferredType>());
+      expect((cfType as ReferredType).declaration.name, equals('C'));
+
+      final egType =
+          getSubTypeOfTypes([e.asReferredType(), g.asReferredType()]);
+      expect(egType, isA<UnionType>());
+      expect((egType as UnionType).types.length, equals(2));
```
srujzs, line 128: I'd check what the types are.

web_generator/test/type_map_test.dart
```diff
@@ -0,0 +1,131 @@
+@TestOn('node')
```
srujzs, line 1: nit: license

web_generator/test/type_map_test.dart
```diff
@@ -0,0 +1,131 @@
...
+
+      final numStringSubType = getSubTypeOfTypes([
+        BuiltinType.primitiveType(PrimitiveType.num),
+        BuiltinType.primitiveType(PrimitiveType.string)
+      ]);
+      expect(numStringSubType.name, equals('JSAny'));
+    });
+
+    test('Sub Type Interface Test', () {
+      final a = InterfaceDeclaration(
```
srujzs, line 68: I'd consider a case where we check the LCA of three types where the pairwise LCAs are different than the LCA of all the types. This is to check we're not accidentally composing the LCAs (which we aren't).

e.g.

```
C -> B -> A
E -> D -> B
F -> H -> B
E -> G -> A
F -> G -> A
```

LCA{E, F} = G but LCA{C, E, F} (B) != LCA{C, LCA{E, F}} (A)

web_generator/lib/src/ast/base.dart
```diff
@@ -85,8 +90,18 @@ abstract class Type extends Node {
   @override
   String? dartName;
 
+  abstract bool isNullable;
```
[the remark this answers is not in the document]
nikeokoronkwo: It isn't uncommon: the whole of `_transformType` (and its fellow functions) depend on this, so I found it much easier setting isNullable than cloning.

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -0,0 +1,337 @@
...
+        outputMap.addAll(createTypeMap([...decl.extendedTypes]));
+      case HomogenousEnumType(types: final homogenousTypes):
+        outputMap.addAll(
+            createTypeMap(homogenousTypes.map((h) => h.baseType).toList()));
+        break;
+      case TupleType(types: final types):
+      case UnionType(types: final types):
+        outputMap.addAll(createTypeMap(types));
+        break;
+      case GenericType(constraint: final constraintedType?):
```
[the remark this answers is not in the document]
nikeokoronkwo: Could you elaborate a bit more?

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -0,0 +1,337 @@
...
+
+  return outputMap;
+}
+
+Type getSubTypeOfTypes(List<Type> types,
+    {bool isNullable = false, TypeMap? typeMap}) {
+  typeMap ??= createTypeMap(types);
+
+  if (types.isEmpty) throw Exception('You must pass types');
+  if (types.singleOrNull case final singleType?) {
```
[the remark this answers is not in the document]
nikeokoronkwo: What if the type is not single (i.e more than one)?

web_generator/lib/src/ast/declarations.dart
```diff
@@ -602,16 +628,17 @@ class InterfaceDeclaration extends TypeDeclaration {
       super.methods,
       super.properties,
       super.operators,
-      super.constructors})
+      super.constructors,
+      this.assertRepType = false,
```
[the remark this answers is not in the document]
nikeokoronkwo: It is used in `FunctionType`

web_generator/lib/src/ast/helpers.dart
```diff
@@ -133,3 +142,170 @@ Type getClassRepresentationType(ClassDeclaration cl) {
...
+          .fold(<GenericType>[], (prev, combine) => [...prev, ...combine]);
+
+      types.addAll(typeParams.where((t) {
+        return alreadyEstablishedTypeParams.any((al) => al.name == t.name);
+      }).map((t) => (t.name, t.constraint)));
+      break;
+    default:
+      break;
+  }
+  return types.map((t) => GenericType(name: t.$1, constraint: t.$2)).toList();
```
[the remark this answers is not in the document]
nikeokoronkwo: I tried doing that before, but modifications to constraints broke a lot of things concerning code generation because of referencing (i.e having `T extends _i1.Any get foo` instead of `T get foo`), so I went in favour of cloning, so that the modification of constraints is done by the declaration where appropriate.

web_generator/lib/src/ast/helpers.dart
```diff
@@ -133,3 +142,170 @@ Type getClassRepresentationType(ClassDeclaration cl) {
...
+}
+
+Type getDeepType(Type t) {
```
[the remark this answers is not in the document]
nikeokoronkwo: I'll hopefully get better at naming things

web_generator/lib/src/ast/helpers.dart
```diff
@@ -133,3 +142,170 @@ Type getClassRepresentationType(ClassDeclaration cl) {
...
+  if (t case final ReferredType ref
+      when ref.declaration is TypeAliasDeclaration) {
+    return getDeepType((ref.declaration as TypeAliasDeclaration).type);
+  }
+  return t;
+}
+
+Declaration generateTupleDeclaration(int count, {bool readonly = false}) {
```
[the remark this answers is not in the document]
nikeokoronkwo: The declaration is generated on demand, since all instances of a tuple are basically `JSTuple` with generics, so they are not paired to a type unfortunately.

web_generator/lib/src/ast/types.dart
```diff
@@ -196,3 +273,326 @@ enum LiteralKind {
...
+  @override
+  bool isNullable;
+
+  @override
+  final String declarationName;
+
+  @override
+  final ID id;
+
+  @override
```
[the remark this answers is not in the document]
nikeokoronkwo: Could you please elaborate? anonymous declarations do not have names, so I put the getter as `null` therefore

web_generator/lib/src/interop_gen/transform.dart
```diff
@@ -61,13 +62,18 @@ class TransformResult {
...
             'constant_identifier_names',
             'non_constant_identifier_names',
             if (declMap.values
                 .any((d) => d is NestableDeclaration && d.parent != null))
               'camel_case_types',
-          ])
+            if (declMap.values.any((v) => v.id.name.contains('Anonymous'))) ...[
+              'camel_case_types',
+              'library_private_types_in_public_api',
+              'unnecessary_parenthesis'
```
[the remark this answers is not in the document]
nikeokoronkwo: Yes

web_generator/test/integration/interop_gen/namespaces_expected.dart
nikeokoronkwo: Does that mean the anonymous union declarations should be non-private?

web_generator/test/integration/interop_gen/ts_typing_expected.dart
nikeokoronkwo: This is an issue I've been thinking about for a while, for reasons you've mentioned and also because of the issue where `JSTuple` in `a.dart` is different from `JSTuple` in `b.dart`. I wanted to create a new file for these, but haven't gotten around to how the implementation will look like. 
We can add these declarations to common declarations for every program map so they persist between source code transformers.
What will the file be called? 

Do you feel we should work on this in this PR: I was considering doing this when getting around to core types in [#380](https://github.com/dart-lang/web/issues/380).

web_generator/test/integration/interop_gen/ts_typing_expected.dart
nikeokoronkwo: - The constructor call shouldn't be a getter, it should be an override on `call`, so it is intended to be called as `productConstr()`. 
- I do not know the implications of having it as a variable vs as a function.
- I feel the anonymous extensions are useful, as in this example, they can be extended/implemented and typealiased.

Do you have any thoughts concerning these?

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -0,0 +1,337 @@
...
+    }
+  }
+
+  void add(TypeHierarchalTree map) {
+    map.parent = this;
+    return nodes.add(map);
+  }
+
+  void addValues(Iterable<String> list) {
+    if (list.isEmpty) {
```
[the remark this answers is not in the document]
nikeokoronkwo: It is meant to add the values as a straight line (node-edge chaining)

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  ({int level, List<int> path})? _lookup(
+      String value, int level, List<int> indexPath) {
+    if (this.value == value) return (level: level, path: indexPath);
+    if (nodes.isEmpty) {
+      return null;
+    } else {
+      // find value
+      return nodes.indexed
```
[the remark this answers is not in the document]
nikeokoronkwo: Talked over this offline, but we should refactor this, as we can have multiple paths

web_generator/lib/src/ast/builtin.dart
```diff
@@ -22,19 +22,14 @@ class BuiltinType extends Type {
   final bool fromDartJSInterop;
 
   @override
-  bool get isNullable => _isNullable ?? false;
-
-  final bool? _isNullable;
-
-  @override
-  set isNullable(bool isNullable) {}
+  bool isNullable;
```
srujzs, line 25: Can we not make this `late` and only set it when we have the value instead of defaulting like we do in line 32?

web_generator/lib/src/ast/builtin.dart
```diff
@@ -51,7 +46,7 @@ class BuiltinType extends Type {
           .where((p) => typeParams.length != 1 || p != $voidType)
           .map((p) => p.emit(options)))
       ..url = fromDartJSInterop ? 'dart:js_interop' : null
-      ..isNullable = _isNullable ?? options?.nullable);
+      ..isNullable = isNullable || (options?.nullable ?? false));
```
srujzs, line 49: In theory, `isNullable` should have been set so we should not see any late initialization errors with the propose change above.

web_generator/lib/src/ast/helpers.dart
```diff
@@ -96,11 +97,19 @@ Set<String> getMemberHierarchy(TypeDeclaration type,
...
     return switch (extendee) {
-      final ClassDeclaration classExtendee =>
-        getClassRepresentationType(classExtendee),
+      ReferredType(declaration: final d) when d is ClassDeclaration =>
+        getClassRepresentationType(d),
+      final BuiltinType b => b,
+      _ => BuiltinType.primitiveType(PrimitiveType.object, isNullable: false)
+    };
+  } else if (cl case InterfaceDeclaration(extendedTypes: [final extendee])) {
+    return switch (extendee) {
```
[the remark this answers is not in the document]
srujzs: I guess the same question applies when `cl` (which should maybe be called `td` now :D) is a `ClassDeclaration` as well.

web_generator/lib/src/ast/helpers.dart
```diff
@@ -173,36 +174,44 @@ List<GenericType> getGenericTypes(Type t) {
...
+          final genericTypes = getGenericTypes(type);
+          for (final genericType in genericTypes) {
+            if (alreadyEstablishedTypeParams
+                .any((al) => al.name == genericType.name)) {
+              types.add((genericType.name, genericType.constraint));
+            }
+          }
+        }
       }
       break;
```
srujzs, line 216: nit: On lines 222-228, avoid the fold similar to above.

web_generator/lib/src/ast/helpers.dart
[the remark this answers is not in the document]
srujzs: Interesting, I'd leave a quick comment mentioning that because I'll likely forget about this convo and might accidentally try to do the same thing. :)

web_generator/lib/src/ast/types.dart
```diff
@@ -517,12 +518,16 @@ class _UnionDeclaration extends NamedDeclaration
       this.types = const [],
       this.isNullable = false,
       List<GenericType>? typeParameters})
-      : typeParameters = typeParameters ??
-            types.map(getGenericTypes).fold(<GenericType>[],
-                (prev, combine) => [...prev, ...combine]).map((t) {
+      : typeParameters = typeParameters ?? [] {
```
srujzs, line 521: Note that you're initializing `typeParameters` and this code will execute first so `typeParameters` will never be null, and we'll never run the body of the `if` condition. Do you mean to check if it's empty instead?

web_generator/lib/src/interop_gen/sub_type.dart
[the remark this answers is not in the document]
srujzs: Only a little bit: `constrainted` isn't a word. :)

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -227,63 +318,35 @@ Type getSubTypeOfTypes(List<Type> types,
...
-    final lookupResult = typeMaps.first.lookup(type)!;
-    final typeMap = typeMaps.first.getMapWithLookup(lookupResult.path);
-
-    // remove types in map
-    markedForRemoval.addAll(commonTypes.where(
-        (t) => typeMap.lookup(t) != null && typeMap.lookup(t)!.level != 0));
+  final topoList = topologicalList(typeMaps.toList());
+  for (final level in topoList) {
+    final typesAtLevel = commonTypes.intersection(level);
+    if (typesAtLevel.isNotEmpty) {
```
srujzs, line 330: I'd just briefly leave a comment here mentioning that by definition, another least common ancestor can't be in the next level.

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -56,19 +61,14 @@ class TypeHierarchalTree {
...
     } else if (list.length == 1) {
-      nodes.add(TypeHierarchalTree(list.single));
+      nodes.add(TypeHierarchy(list.single));
       return;
     }
-    nodes.add(TypeHierarchalTree(list.first)..addValues(list.skip(1)));
+    nodes.add(TypeHierarchy(list.first)..addChainedValues(list.skip(1)));
   }
 
   Set<String> expand() {
```
srujzs, line 74: I'd cache in this method as we use the result of this for intersections.

srujzs (approved):
Thanks, Nike! Just a couple more comments.

web_generator/lib/src/ast/builtin.dart
nikeokoronkwo: We can, but it isn't guaranteed to be set. It is only set if `isNullable` is set

```dart
/// isNullable can be null
Type _transformType(TSTypeNode type,
      {bool parameter = false, bool typeArg = false, bool? isNullable}) {
        // code 
      }
```

web_generator/lib/src/ast/helpers.dart
```diff
@@ -133,3 +142,170 @@ Type getClassRepresentationType(ClassDeclaration cl) {
...
+              alreadyEstablishedTypeParams.any((al) => al.name == t.name);
+        }).map((t) => (t.name, t.constraint)));
+      }
+      break;
+    case ClosureType(
+        typeParameters: final alreadyEstablishedTypeParams,
+        returnType: final closureType,
+        parameters: final closureParams
+      ):
+      final typeParams = [closureType, ...closureParams.map((p) => p.type)]
```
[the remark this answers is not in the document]
nikeokoronkwo: Oh missed this one!

web_generator/lib/src/ast/helpers.dart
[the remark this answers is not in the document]
srujzs: Sorry, I meant having this method be a factory in the class `_TupleDeclaration` instead of a top-level definition here.

web_generator/lib/src/ast/types.dart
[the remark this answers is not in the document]
srujzs: Right that's what I'm referring to when I say `name` maybe doesn't make sense for all `DeclarationType`s as there are multiple subtypes that don't have a `name`.

web_generator/test/integration/interop_gen/ts_typing_expected.dart
srujzs: I think it's fine to throw into this PR. We could simply call it `tuples.dart` and have it be included in every output that uses tuples.

web_generator/lib/src/interop_gen/sub_type.dart
nikeokoronkwo: > Under what circumstances is nodes modified?

It shouldn't be in our current algorithm.

web_generator/lib/src/ast/base.dart
```diff
@@ -147,3 +145,7 @@ class ParameterDeclaration {
       ..type = type.emit(TypeOptions(nullable: optional)));
   }
 }
+
+abstract class NamedType extends Type {
```
srujzs, line 149: Awesome! I didn't expect there would be such an elegant fix to the `null` `name`s.

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -192,6 +215,7 @@ TypeHierarchy getTypeHierarchy(Type type) {
           hierarchy.nodes.add(getTypeHierarchy(
               BuiltinType.primitiveType(PrimitiveType.object)));
         } else {
+          print((decl.name, decl.extendedTypes));
```
srujzs, line 218: nit: stray print

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -314,26 +346,32 @@ Type getLowestCommonAncestorOfTypes(List<Type> types,
...
-  if (_getSharedTypeIfAny(types, isNullable: isNullable) case final t?) {
+  if (_getSharedPrimitiveTypeIfAny(types, isNullable: isNullable)
+      case final t?) {
     return t;
   }
 
   // Calculate the intersection of all type hierarchies
   final typeMaps = types.map(getTypeHierarchy);
   final parentHierarchy = typeMaps.map((map) => map.expand());
+  print(parentHierarchy);
```
srujzs, line 357: nit: stray print

web_generator/lib/src/interop_gen/sub_type.dart
```diff
@@ -314,26 +346,32 @@ Type getLowestCommonAncestorOfTypes(List<Type> types,
...
   final commonTypes =
       parentHierarchy.reduce((val, element) => val.intersection(element));
 
+  print(commonTypes);
```
srujzs, line 361: nit: stray print

web_generator/test/integration/interop_gen/ts_typing_expected.dart
srujzs: > In either case, if a typedef is placed, it'd be useful to have it as a type to typedef as, wouldn't it?

*Maybe*. :) But I'm fine leaving it as is.

srujzs (approved):
Looks good! Just some small nits around prints.

web_generator/test/integration/interop_gen/web_types_expected.dart
```diff
@@ -72,88 +73,217 @@ extension type HTMLTransformFunc<T extends _i2.HTMLElement,
...
+
+  /// Returns a date as a string value appropriate to the host environment's current locale.
+  /// Converts a date to a string by using the current or specified locale.
+  /// - [locales]:  A locale string or array of locale strings that contain one or more language or locale tags. If you include more than one locale string, list them in descending order of priority so that the first entry is the preferred locale. If you omit this parameter, the default locale of the JavaScript runtime is used.
+  /// - [options]:  An object that contains one or more properties that specify comparison options.
   external String toLocaleDateString();
+
+  /// Returns a time as a string value appropriate to the host environment's current locale.
+  /// Converts a time to a string by using the current or specified locale.
+  /// - [locales]:  A locale string or array of locale strings that contain one or more language or locale tags. If you include more than one locale string, list them in descending order of priority so that the first entry is the preferred locale. If you omit this parameter, the default locale of the JavaScript runtime is used.
```
srujzs, line 103: Can we wrap strings that are greater than the line length? See `formatter.dart` where we do this exact thing.

PR #444, opened 2025-08-12
[web] update output w/ latest generator

kevmoo:


srujzs (commented):
@nikeokoronkwo Do we need this lint for the web package or can it be only for TS gen?

srujzs:
Right, but we already ignore them for the whole `web` package in the analysis_options.yaml file so this is duplicating that: https://github.com/dart-lang/web/blob/4310354266abaffae08fdb483327a8523585aed4/web/analysis_options.yaml#L22.

kevmoo:
It might be better to do it file-by-file and drop the pkg:web analyzer bit. Just to keep things consistent .

nikeokoronkwo:
> Yeah, it's a bit odd because we're already inconsistent. We have other ignores in that analysis_options.yaml file but we have an ignore in each of the files for `unintended_html_in_doc_comment`.

I think we can keep that, since we wouldn't necessarily be testing single-file output with already-established web APIs (that'd be suited for something like the current repo)

srujzs:
> I think we can keep that

You mean keep ignores for `unintended_html_in_doc_comment` in each of the files?

At any rate, if we want this per file, that's fine by me, but we should just remove the global ignore in analysis_options.yaml to avoid duplication.

nikeokoronkwo:
We might keep it per-file for now, because there's a setting to auto-generate docs for any decl that is part of webref. In the future, we can have such setting an option only enabled for generating `package:web` and then move to the global `analysis_options.yaml` file

PR #476, opened 2025-10-15
Initial commit of package:js_interop 0.1.0-beta

srujzs:
Making a breaking change to avoid breaking any existing users of package:js_interop. Also modifies the monorepo's README to detail the packages within rather than being a symlink to web's README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,4 @@
-{% comment %}
-TODO(srujzs): Add links to CI once we have a badge that doesn't say "package:web" and is more generic.
-{% endcomment %}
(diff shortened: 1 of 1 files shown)


kevmoo (approved):
Yes!

devoncarew:
We may consider refactoring this mono-repo so that its packages are in a `pkgs/` sub-directory. That would mirror our other mono-repos (https://github.com/dart-lang/core, https://github.com/dart-lang/tools, ...), and would play better w/ some of our home-grown tools.

Concretely I think that would look like:

- `js_interop` => `pkgs/js_interop`
- `web` => `pkgs/web`
- `web_generator` => either `tool/web_generator` or `pkgs/web_generator`

Thoughts?

Issue #479, opened 2025-10-30
Regenerating pkg:web adds a bunch of extra lint ignores


kevmoo:
```diff
diff --git a/web/lib/src/dom/accelerometer.dart b/web/lib/src/dom/accelerometer.dart
index dc117c0..984268f 100644
--- a/web/lib/src/dom/accelerometer.dart
+++ b/web/lib/src/dom/accelerometer.dart
@@ -8,7 +8,7 @@

 // Generated from Web IDL definitions.

-// ignore_for_file: unintended_html_in_doc_comment
+// ignore_for_file: constant_identifier_names, unintended_html_in_doc_comment

 @JS()
 library;
```

Should that be fixed?

PR #481, opened 2025-10-30
Modify lint ignores in package:web

srujzs:
Fixes https://github.com/dart-lang/web/issues/479

Now that we have a general purpose IDL->interop generator, we need to add ignores for constant_identifier_names and non_constant_identifier_names in every output file. unintended_html_in_doc_comment no longer applies as the original reason for it has been fixed.

Removes the two lints from analysis_options.yaml as it would now be redundant.


kevmoo (approved):
LGTM. Good luck w/ the CI weirdness. 😄

Issue #480, opened 2025-10-30
Crash when running `dart web_generator/bin/update_idl_bindings.dart --update`


kevmoo:
```
dart web_generator/bin/update_idl_bindings.dart --update
* npm update

changed 6 packages, and audited 10 packages in 2s

found 0 vulnerabilities
* /Users/kevmoo/github/flutter/bin/cache/dart-sdk/bin/dart compile js --enable-asserts --server-mode -DlanguageVersion=3.4.0 dart_main.dart -o dart_main.js
Hint: When run on the command-line, the compiled output might require a preamble file located in:
  <sdk>/lib/_internal/js_runtime/lib/preambles.
Compiled 18,717,388 input bytes (15,841,674 characters source) to 3,860,581 characters JavaScript in 5.96 seconds
* node main.mjs --idl --output=/Users/kevmoo/github/web/web/lib/src
/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:5795
      throw error;
      ^

Exception: Unrecognized member type async_iterable
    at Object.wrapException (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:1480:47)
    at _PartialInterfacelike._processMembers$1 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83756:23)
    at Translator._combineInterfacelikes$1 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83882:32)
    at Translator.markTypeAsUsed$1 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83934:19)
    at Object._getRawType (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:20711:29)
    at _PartialInterfacelike._processMembers$1 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83722:24)
    at Translator._combineInterfacelikes$1 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83882:32)
    at Translator.markTypeAsUsed$1 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83934:19)
    at Translator.addInterfacesAndNamespaces$0 (/Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:83837:20)
    at /Users/kevmoo/github/web/web_generator/lib/src/dart_main.js:19562:26 {
  dartException: _Exception { message: 'Unrecognized member type async_iterable' }
}

Node.js v25.1.0
[...]

PR #482, opened 2025-10-30
Update web library and ignore async_iterables

srujzs:
Fixes https://github.com/dart-lang/web/issues/480

async_iterables are a new type of member added to the Web IDL. Currently, we don't have an iterator type to produce anything useful so ignore for now.

Updates the constraint for the compat data to use the latest.


srujzs:
Sorry for the delay, this led to some TS gen tests failing: https://github.com/dart-lang/web/issues/494. For now, I'll pin the dependency and this should be ready.

kevmoo:
@srujzs – I did the merge for you

PR #484, opened 2025-11-06
Add typings for JSDate

nex3:
- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/js_interop/test/date_test.dart
+++ b/js_interop/test/date_test.dart
@@ -1,0 +1,212 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.p
+
+import 'dart:js_interop';
+
+import 'package:test/test.dart';
+
+import 'package:js_interop/js_interop.dart';
+
+final Matcher _isDate = predicate(
+  (jsDate) => jsDate is JSObject && jsDate.isA<JSDate>(),
+  "is a JSDate",
+);
+
+void main() {
+  group('constructors', () {
+    test('nowAsDate', () => expect(JSDate.nowAsDate(), _isDate));
+
+    test(
+      'fromMillisecondsSinceEpoch',
+      () => expect(JSDate.fromMillisecondsSinceEpoch(1234567890), _isDate),
+    );
+
+    test(
+      'parseAsDate',
+      () => expect(JSDate.parseAsDate("11 Nov 2025"), _isDate),
+    );
+
+    test('from', () => expect(JSDate.from(JSDate.nowAsDate()), _isDate));
+
+    group('localDate', () {
+      test(
+        'with no optional args',
+        () => expect(JSDate.localDate(2025, 11), _isDate),
+      );
+
+      test(
+        'with all optional args',
+        () => expect(JSDate.localDate(2025, 11, 11, 12, 2, 15, 120), _isDate),
+      );
+    });
+
+    group('utcDate', () {
+      test(
+        'with no optional args',
+        () => expect(JSDate.utcDate(2025, 11), _isDate),
+      );
+
+      test(
+        'with one optional arg',
+        () => expect(JSDate.utcDate(2025, 11, 11), _isDate),
+      );
+
+      test(
+        'with two optional args',
+        () => expect(JSDate.utcDate(2025, 11, 11, 12), _isDate),
+      );
+
+      test(
+        'with three optional args',
+        () => expect(JSDate.utcDate(2025, 11, 11, 12, 2), _isDate),
+      );
+
+      test(
+        'with four optional arg',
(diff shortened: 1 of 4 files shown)


kevmoo:
No tests? 😁

js_interop/lib/src/date.dart
```diff
@@ -0,0 +1,314 @@
...
+  external JSDate.from(JSDate other);
+
+  /// The [Date constructor] that uses [individual component integers].
+  ///
+  /// [Date constructor]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date
+  /// [individual component integers]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#individual_date_and_time_component_values
+  external JSDate(int year, int month,
+      [int? day, int? hours, int? minutes, int? seconds, int? milliseconds]);
+
+  /// Dee [`Date.now()`].
```
srujzs: nit: `See`

js_interop/lib/src/date.dart
```diff
@@ -0,0 +1,314 @@
...
+  ///
+  /// [`Date.parse()`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse
+  @JS('parse')
+  external static int parseAsMillisecondsSinceEpoch(String dateString);
+
+  /// The [`Date.utc()`] function.
+  ///
+  /// [`Date.utc()`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC
+  @JS('UTC')
+  external static int utc(int year,
```
srujzs: We avoid renaming to lowercase with `package:web` e.g. `URL`, maybe worth doing the same?

Same comment on `toIsoString`, `toUtcString`, and `toDartUtc`.

srujzs:
> Authors will never be able to rely on names being JS-style without caveats, but they would be able to rely on them being Dart-style without caveats.

I agree that there will always be exceptions due to the nature of keywords/Dart semantics, and I can see the argument that e.g. reaching for the all-uppercase version is less intuitive than the camel-cased version, but I think a strict adherence to Dart style for interop, e.g. `getHours` and `setHours` being moved to a getter-setter pair, can be less intuitive and requires reading the API documentation. Maybe my mental model is wrong, but I assume interop users start from wanting to call a JS member/functionality and work their way backwards.

> I feel more strongly about this for places where the JS names outright violate Dart naming conventions (like Date.utc) than I do for cases where I'm reusing a JS name to do something slightly different (like Date.now). Maybe we can compromise on only doing the former and renaming the latter to Date.nowAsDate and so on?

Sounds good, I don't have a strong opinion on capitalization. `utc` should be barely less findable than `UTC`.

js_interop/lib/src/date.dart
```diff
@@ -0,0 +1,375 @@
...
+  factory JSDate.utcDate(
+    int year,
+    int month, [
+    int? day,
+    int? hours,
+    int? minutes,
+    int? seconds,
+    int? milliseconds,
+  ]) {
+    var ms = switch ((day, hours, minutes, seconds, milliseconds)) {
```
kevmoo: Love me some switch!

js_interop/test/date_test.dart
```diff
@@ -0,0 +1,212 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.p
+
+import 'dart:js_interop';
+
+import 'package:test/test.dart';
+
+import 'package:js_interop/js_interop.dart';
+
+final Matcher _isDate = predicate(
```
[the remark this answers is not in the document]
kevmoo: Gah! Right. Ugh.

srujzs:
> I wonder if this is the core disconnect here.

Probably. :D I also think this will differ between folks who will use these types regularly versus folks who use these once.

> If I'm writing JS interop code in Dart, I'd be looking those up in the Dart documentation anyway, so having them be a little more different from the JS in exchange for them being a little more ergonomic at every use-site is a tradeoff that's well worth its cost to me.

I suppose one benefit types in this package will have over `package:web` is they'll be more intuitive and require less reason to go look for documentation if the name makes it fairly obvious.

----

But okay, so that we make progress, let's go with the strategy that you propose and make things more Dart-y and ergonomic where it makes sense. In this PR that means capitalization modifications and moving obvious methods to getters/setters.

I expect us to do a pre-release, and we can get more input (if any) from other users on the conventions then before releasing a 1.0.0.

PR #487, opened 2025-12-10
Add typings for JSRecord and some unsafe extensions for JSObject

nex3:
This doesn't include JSObject extensions that require types that
aren't defined yet, like JSSymbolicRecord.

### DIFF
--- a/js_interop/lib/src/record.dart
+++ b/js_interop/lib/src/record.dart
@@ -66,7 +66,7 @@ extension type JSRecord<V extends JSAny?>._(JSObject _) implements JSObject {
 
   /// Creates a [JSRecord] with the same keys and values as [other].
   static JSRecord<V> ofRecord<V extends JSAny?>(JSRecord<V> other) =>
-      JSRecord<V>()..addAllRecord(other);
+      JSRecord<V>()..addAllFromRecord(other);
 
   /// Like [Map.of], but creates a record.
   static JSRecord<V> ofMap<V extends JSAny?>(Map<String, V> other) =>
@@ -95,7 +95,7 @@ extension type JSRecord<V extends JSAny?>._(JSObject _) implements JSObject {
   /// The operation is equivalent to doing `this[key] = value` for each key and
   /// associated value in [other]. It iterates over [other], which must therefore
   /// not change during the iteration.
-  void addAllRecord(JSRecord<V> other) => addPairs(other.pairs);
+  void addAllFromRecord(JSRecord<V> other) => addPairs(other.pairs);
 
   /// See [Map.addEntries].
   void addEntries(Iterable<MapEntry<String, V>> entries) {
--- a/js_interop/lib/src/unsafe/object.dart
+++ b/js_interop/lib/src/unsafe/object.dart
@@ -43,9 +43,8 @@ external bool _set(JSObject object, JSAny name, JSAny? value, JSAny? thisArg);
 @JS('Object.values')
 external JSArray<JSAny?> _values(JSObject object);
 
-/// Additional instance methods for the `dart:js_interop` [interop.JSObject]
-/// type meant to be used when the names of properties or methods are not known
-/// statically.
+/// Additional instance methods for the `dart:js_interop` [JSObject] type meant
+/// to be used when the names of properties or methods are not known statically.
 extension JSObjectUnsafeExtension on JSObject {
   /// See [`Object.entries()`].
   ///
--- a/js_interop/test/record_test.dart
+++ b/js_interop/test/record_test.dart
@@ -92,8 +92,8 @@ void main() {
     );
   });
 
-  test("addAllRecord()", () {
-    record.addAllRecord({"bar": 3.toJS, "baz": 4.toJS}.toJSRecord);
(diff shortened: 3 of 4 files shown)


js_interop/lib/unsafe.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
kevmoo:
```suggestion

/// Something about what this is...
library;
```

js_interop/lib/src/unsafe/object.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  /// See [`Object.assign()`].
+  ///
+  /// [`Object.assign()`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+  void assign([
+    JSObject? source1,
+    JSObject? source2,
+    JSObject? source3,
+    JSObject? source4,
+  ]) => _assign(this, source1, source2, source3, source4);
```
kevmoo: Do we have to worry about `null` vs `undefined` here?

js_interop/lib/src/record.dart
```diff
@@ -0,0 +1,191 @@
...
+
+  /// Adds all enumerable, own, string key/value pairs of [other] to this
+  /// record.
+  ///
+  /// If a key of [other] is already in this record, its value is overwritten.
+  ///
+  /// The operation is equivalent to doing `this[key] = value` for each key and
+  /// associated value in [other]. It iterates over [other], which must therefore
+  /// not change during the iteration.
+  void addAllRecord(JSRecord<V> other) => addPairs(other.pairs);
```
srujzs: tiny nit: maybe `addAllFromRecord` instead?

js_interop/test/record_test.dart
```diff
@@ -0,0 +1,257 @@
...
+
+  test("addPairs()", () {
+    record.addPairs([("bar", 3.toJS), ("baz", 4.toJS)]);
+    expect(
+      record.toDart,
+      equals({"foo": 1.toJS, "bar": 3.toJS, "baz": 4.toJS}),
+    );
+  });
+
+  test("addPairs()", () {
```
srujzs: nit: did you mean to name this as `clear()`?

js_interop/test/unsafe/object_test.dart
```diff
@@ -0,0 +1,68 @@
...
+
+    expect(object.dartify(), equals({"foo": 1, "bar": 2}));
+  });
+
+  test(
+    "getPropertyWithThis",
+    () =>
+        expect(object.getPropertyWithThis("foo".toJS, object), equals(1.toJS)),
+  );
+
```
srujzs: nit: test for `getOwnProperty`

nikeokoronkwo:
> JS objects can't actually have numeric keys; numbers are coerced to strings. 

Agreed, but numeric indexes could also still work, and people/js package authors still rely on this functionality anyways.

The main issue I'm having is that the packages I mentioned use custom types (usually unions, but there are instances of enums as well, which are usually `number`s underlying) as indices, which do reduce to `string` eventually, but at compile time it isn't so clear. Especially with the way dart interop code is generated at the moment, such types, if unions, would show as `JSAny`, and would therefore not fit in such cases (even if we ignored the type, users may want to index with such values rather than having to coerce directly into a `string`/`number`). 

> Symbols are another matter. I plan to bring over another type, JSSymbolicRecord, which uses JSAny as its key type in place of String. I think it's valuable to have access to both, because there are many places where only strings are expected or allowed, and it's much simpler to work with these types when you don't have to manually wrap and unwrap JSString every time. Then we can use JSSymbolicRecord for places where the higher generality is allowed.

In that case, would it be fine if we have `JSRecord` as is (maybe with a warning or something else not to use this directly but to use the child types), and then have a `JSStringRecord` and `JSSymbolicRecord` inheriting from `JSRecord`, which the majority of users would use? That way, packages that don't use strings for indexing can still work, and the majority of people can still use.

js_interop/lib/src/unsafe/object.dart
```diff
@@ -0,0 +1,134 @@
...
+external JSArray<JSAny?> _values(JSObject object);
+
+/// Additional instance methods for the `dart:js_interop` [interop.JSObject]
+/// type meant to be used when the names of properties or methods are not known
+/// statically.
+extension JSObjectUnsafeExtension on JSObject {
+  /// See [`Object.entries()`].
+  ///
+  /// [`Object.entries()`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
+  List<(String, JSAny?)> get entries => [
```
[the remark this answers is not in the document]
srujzs: Okay, I can see the argument.

> in cases where the author knows that the value type is homogeneous, they can now use JSRecord to declare that explicitly.

This is especially convincing, since `pairs` is right there if they wanted a better API. Nothing to do then, LGTM!

PR #488, opened 2025-12-10
Move generator to latest Dart, update analyzer dep

kevmoo:
- add utils so it's easy to update expected output
- fix bug where config language wasn't being passed to the generator in tests
- Reformatted w/ latest SDK

### DIFF
--- a/.github/workflows/web_generator.yaml
+++ b/.github/workflows/web_generator.yaml
@@ -39,7 +39,7 @@ jobs:
 
       - run: dart pub get
       - run: dart format --output=none --set-exit-if-changed .
-        if: matrix.sdk == 'dev' && matrix.test_config == ''
+        if: matrix.test_config == ''
       - run: dart analyze --fatal-infos
         if: matrix.sdk == 'dev' && matrix.test_config == ''
       - run: dart test ${{ matrix.test_config }}


.github/workflows/web_generator.yaml
```diff
@@ -39,7 +39,9 @@ jobs:
 
       - run: dart pub get
       - run: dart format --output=none --set-exit-if-changed .
+        if: matrix.sdk == 'dev' && matrix.test_config == ''
```
kevmoo: Avoid doing format/analyze more than once

web_generator/test/test_shared.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:test/test.dart';
+
+const _rewriteFiles = false;
```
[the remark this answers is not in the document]
srujzs: Why do we need this const?

.github/workflows/web_generator.yaml
kevmoo:
```suggestion
        if: matrix.test_config == ''
```

PR #490, opened 2025-12-11
Mutable `List` interface for `children` and `childNodes`

fsw:
`dart:html` allowed modifying child elements and nodes using List interface:

```dart
import 'dart:html';

document.querySelector('#foo')!.children.removeWhere((e) => e.classes.contains('remove-me'));
```

This was a pretty clear and handy interface IMHO.

As discussed in this PR https://github.com/dart-lang/site-www/pull/7040 with @srujzs it might be useful to implement something similar in `package:web`. This PR is my attempt at doing so.

It adds `JSLiveNodeListWrapper` that requires a container node and a `HTMLCollection` or `NodeList` that is [live](
https://developer.mozilla.org/en-US/docs/Web/API/NodeList#live_vs._static_nodelists) and implements mutable `List` methods.

For easier usage it also adds `childrenAsList` and `childNodesAsList` via extension so you can do:

```dart
import 'package:web/web.dart';

document.querySelector('#foo')!.childrenAsList.removeWhere((e) => e.classList.contains('remove-me'));
document.querySelector('#foo')!.childNodesAsList.removeWhere((e) => e.textContent == 'remove-me');
```

### Questions/Notes:

* Does this implementation make sense and matches ideas behind helpers in this module? `JSLiveNodeListWrapper` don't have to extend `JSImmutableListWrapper` but have its own implementation but it kinda felt it makes sense this way.

* Alternatively handy methods like `removeWhere` could be added via extension directly to `HTMLCollection` and `NodeList` but methods like `add()` might not be possible as those require a parent node to add elements to. This would require knowing if `NodeList` was returned from `querySelector` or `childNodes` and in further case accessing parent somehow. Also implementing full `List` interface might have other pros.

* could/should `NodeList` (returned f.e. from `querySelector`) be extended with `asImmutableList` via extension? I think it would be more similar to above and `dart:html` interfaces and `JSImmutableListWrapper` and `JSLiveNodeListWrapper` could then be used internally or even marked as private.

* While doing this I noticed `ListMixin` and `ListBase` is going to be deprecated in core? Could implementing `List` interface directly by both those wrappers while making changes here give any benefits? 

---

### DIFF
--- a/web/lib/src/helpers/extensions.dart
+++ b/web/lib/src/helpers/extensions.dart
@@ -105,16 +105,6 @@ extension UriToURL on Uri {
   }
 }
 
-extension NodeExtension on Node {
-  /// Returns [childNodes] as a modifiable [List].
-  List<Node> get childNodesAsList => NodeListListWrapper(this, childNodes);
-}
-
-extension ElementExtension on Element {
-  /// Returns [children] as a modifiable [List].
-  List<Element> get childrenAsList => HTMLCollectionListWrapper(this, children);
-}
-
 extension NodeListExtension on NodeList {
   /// Returns node list as a modifiable [List].
   List<Element> get asList => JSImmutableListWrapper(this);
--- a/web/lib/src/helpers/lists.dart
+++ b/web/lib/src/helpers/lists.dart
@@ -159,26 +159,17 @@ abstract mixin class _LiveNodeListMixin<P extends Node, U extends Node> {
     _filter(test, true);
   }
 
-  Iterator<U> get iterator;
-
   void _filter(bool Function(U element) test, bool requiredTestValue) {
-    // This implementation of removeWhere/retainWhere is more efficient
-    // than the default in ListBase. Child nodes can be removed in constant
-    // time.
-    final i = iterator;
-    U? removeMe;
-    while (i.moveNext()) {
-      if (removeMe != null) {
-        _parent.removeChild(removeMe);
-        removeMe = null;
-      }
-      if (test(i.current) != requiredTestValue) {
-        removeMe = i.current;
+    final toRemove = <U>[];
+    final length = this.length;
+    for (var i = 0; i < length; i++) {
+      final element = this[i];
+      if (test(element) != requiredTestValue) {
+        toRemove.add(element);
       }
     }
-    if (removeMe != null) {
-      _parent.removeChild(removeMe);
-      removeMe = null;
+    for (var element in toRemove) {
+      _parent.removeChild(element);
     }
   }
 
@@ -233,30 +224,43 @@ abstract mixin class _LiveNodeListMixin<P extends Node, U extends Node> {
   }
 }
 
-/// Allows iterating `HTMLCollection` with `nextElementSibling` for optimisation and easier encapsulation
-class _HTMLCollectionIterator implements Iterator<Element> {
+/// Iterates [_LiveNodeListMixin]. Compared to default `ListIterator`, this
+/// iterator skips `length` check for `ConcurrentModificationError` to improve
+/// performance in WASM.
+class _NodeListIterator<E> implements Iterator<E> {
+  final Iterable<E> _iterable;
+  final int _length;
+  int _index;
+  E? _current;
+
+  _NodeListIterator(Iterable<E> iterable)
+      : _iterable = iterable,
+        _length = iterable.length,
+        _index = 0;
+
   @override
-  Element get current => _current!;
-
-  Element? _current;
-  bool start = true;
-
-  _HTMLCollectionIterator(this._current);
+  E get current => _current!;
 
   @override
+  @pragma('vm:prefer-inline')
+  @pragma('wasm:prefer-inline')
   bool moveNext() {
-    if (start) {
-      start = false;
+    if (_index >= _length) {
+      _current = null;
+      return false;
     } else {
-      _current = _current?.nextElementSibling;
+      _current = _iterable.elementAt(_index);
+      _index++;
+      return true;
     }
-    return _current != null;
   }
 }
 
-/// Wrapper for `HTMLCollection` returned from `children` that implements modifiable list interface and allows easier DOM manipulation.
-/// This is loosely based on `_ChildrenElementList` from `dart:html` to preserve compatibility
-class HTMLCollectionListWrapper
+/// Wrapper for `HTMLCollection` returned from `children` that implements
+/// modifiable list interface and allows easier DOM manipulation.
+/// This is loosely based on `_ChildrenElementList` from `dart:html` to
+/// preserve compatibility.
+class _HTMLCollectionListWrapper
     with ListMixin<Element>, _LiveNodeListMixin<Element, Element> {
   @override
   final Element _parent;
@@ -265,11 +269,12 @@ class HTMLCollectionListWrapper
 
   final HTMLCollection _htmlCollection;
 
-  HTMLCollectionListWrapper(this._parent, this._htmlCollection);
+  _HTMLCollectionListWrapper(this._parent, this._htmlCollection);
 
   @override
-  Iterator<Element> get iterator =>
-      _HTMLCollectionIterator(_parent.firstElementChild);
+
+  ///See [_NodeListIterator] for information.
+  Iterator<Element> get iterator => _NodeListIterator(this);
 
   @override
   bool get isEmpty {
@@ -306,30 +311,10 @@ class HTMLCollectionListWrapper
   }
 }
 
-/// Allows iterating `NodeList` with `nextSibling` for optimisation and easier encapsulation
-class _NodeListIterator implements Iterator<Node> {
-  @override
-  Node get current => _current!;
-
-  Node? _current;
-  bool start = true;
-
-  _NodeListIterator(this._current);
-
-  @override
-  bool moveNext() {
-    if (start) {
-      start = false;
-    } else {
-      _current = _current?.nextSibling;
-    }
-    return _current != null;
-  }
-}
-
 /// Wrapper for `NodeList` returned from `childNodes` that implements modifiable list interface and allows easier DOM manipulation.
 /// This is loosely based on `_ChildNodeListLazy` from `dart:html` to preserve compatibility
-class NodeListListWrapper with ListMixin<Node>, _LiveNodeListMixin<Node, Node> {
+class _NodeListListWrapper
+    with ListMixin<Node>, _LiveNodeListMixin<Node, Node> {
   @override
   final Node _parent;
   @override
@@ -337,10 +322,12 @@ class NodeListListWrapper with ListMixin<Node>, _LiveNodeListMixin<Node, Node> {
 
   final NodeList _nodeList;
 
-  NodeListListWrapper(this._parent, this._nodeList);
+  _NodeListListWrapper(this._parent, this._nodeList);
 
   @override
-  Iterator<Node> get iterator => _NodeListIterator(_parent.firstChild);
+
+  ///See [_NodeListIterator] for information.
+  Iterator<Node> get iterator => _NodeListIterator(this);
 
   @override
   bool get isEmpty {
@@ -375,4 +362,17 @@ class NodeListListWrapper with ListMixin<Node>, _LiveNodeListMixin<Node, Node> {
       _parent.removeChild(_parent.firstChild!);
     }
   }
+}
+
+extension NodeExtension on Node {
+  /// Returns [childNodes] as a modifiable [List].
+  /// This replaces functionality of the `nodes` getter from `dart:html`.
+  List<Node> get childNodesAsList => _NodeListListWrapper(this, childNodes);
+}
+
+extension ElementExtension on Element {
+  /// Returns [children] as a modifiable [List].
+  /// This replaces functionality of the `children` getter from `dart:html`.
+  List<Element> get childrenAsList =>
+      _HTMLCollectionListWrapper(this, children);
 }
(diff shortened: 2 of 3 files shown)


web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,35 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+  @override
+  void operator []=(int index, Node value) {
+    parentNode.replaceChild(value, _jsList.item(index));
+  }
+
+  @override
+  void add(Node element) {
+    // `ListMixin` implementation only works for lists that allow `null`.
+    parentNode.appendChild(element);
+  }
```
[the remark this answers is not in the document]
fsw: fixed, thanks ai.

web/test/helpers_test.dart
```diff
@@ -40,6 +40,55 @@ void main() {
...
+    // replace element with text node
+    childNodesList[2] = document.createTextNode('txt3');
+
+    // test retainWhere, keep Elements only
+    childNodesList.retainWhere((e) => e.isA<Element>());
+
+    // Ensure only text nodes were removed
+    expect(childNodesList.length, 4);
+    expect(childrenList.length, 4);
+  });
```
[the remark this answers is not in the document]
fsw: fixed, thanks ai

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,44 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
   U elementAt(int index) => this[index];
 }
+
+/// A wrapper for live node lists. `NodeList` and `HTMLCollection` that are
+/// [live](https://developer.mozilla.org/en-US/docs/Web/API/NodeList#live_vs._static_nodelists)
+/// can be safely modified at runtime. This requires an instance of `P`, a
+/// container that elements would be added to or removed from.
+class JSLiveNodeListWrapper<P extends Node, T extends JSObject, U extends Node>
+    extends JSImmutableListWrapper<T, U> {
+  P parentNode;
```
kevmoo: this could/should be final?


```suggestion
  final P parentNode;
```

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,35 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+  void operator []=(int index, Node value) {
+    parentNode.replaceChild(value, _jsList.item(index));
+  }
+
+  @override
+  void add(Node element) {
+    // `ListMixin` implementation only works for lists that allow `null`.
+    parentNode.appendChild(element);
+  }
+}
```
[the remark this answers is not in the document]
fsw: It was right about this being not optimal (without this override `replaceChild` and `removeChild` would be called N times) but indeed its proposed solution was also not optimal as you pointed out. So i fixed it by doing this instead:

```dart
  @override
  void removeRange(int start, int end) {
    RangeError.checkValidRange(start, end, length);
    for (var i = 0; i < end - start + 1; i++) {
      parentNode.removeChild(this[start]);
    }
  }
```
And now only `removeChild`  is called N times.

This might be an issue with other methods and might be related to my question about `ListMixin` (should we use it or implement `List`).

fsw:
> * `ListMixin` is not deprecated and is the correct tool for this job.

For clarification, I am talking about this comments:

https://github.com/dart-lang/sdk/blob/9a049d5b2defab33aa953e7882492639ce1f677b/sdk/lib/collection/list.dart#L33
https://github.com/dart-lang/sdk/blob/9a049d5b2defab33aa953e7882492639ce1f677b/sdk/lib/collection/list.dart#L585

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,44 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+  final P parentNode;
+
+  JSLiveNodeListWrapper(this.parentNode, super.original);
+
+  @override
+  set length(int value) {
+    if (value > length) {
+      throw UnsupportedError('Cannot add null to live node List.');
+    }
+    for (var i = length - 1; i >= value; i--) {
```
srujzs: I believe this is new compared to the dart:html version. I think that's okay though since we were throwing in the dart:html version anyways.

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,44 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+  @override
+  void add(U element) {
+    // `ListMixin` implementation only works for lists that allow `null`.
+    parentNode.appendChild(element);
+  }
+
+  @override
+  void removeRange(int start, int end) {
+    RangeError.checkValidRange(start, end, length);
+    for (var i = 0; i < end - start + 1; i++) {
```
srujzs: This is also new compared to the dart:html version, but again, it was throwing before so this is okay.

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,44 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+      parentNode.removeChild(this[start]);
```
srujzs: Should this be `start + i`?

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,44 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
   @override
   U elementAt(int index) => this[index];
 }
+
+/// A wrapper for live node lists. `NodeList` and `HTMLCollection` that are
+/// [live](https://developer.mozilla.org/en-US/docs/Web/API/NodeList#live_vs._static_nodelists)
+/// can be safely modified at runtime. This requires an instance of `P`, a
+/// container that elements would be added to or removed from.
+class JSLiveNodeListWrapper<P extends Node, T extends JSObject, U extends Node>
```
srujzs: `dart:html` has overrides for `first` and `last` for `children` that use `firstElementChild` and `lastElementChild`, respectively.

This I believe is different from `item(0)` and `item(len - 1)`, so I wonder if we should split the implementation here so `childNodes` and `children` work differently.

----

As an aside, `remove` is also overridden in `dart:html` and uses `identical` (which won't work with `dart:js_interop` types and would need `===`/`strictEquals` from `dart:js_interop`). I think the default implementation in `ListMixin` that uses `==` is okay as I don't believe the differences between `==` and `===` can materialize in `children` or `childNodes`.

web/lib/src/helpers/extensions.dart
```diff
@@ -103,3 +104,14 @@ extension UriToURL on Uri {
     }
   }
 }
+
+extension NodeExtension on Node {
+  /// Returns [childNodes] as a modifiable [List]
+  List<Node> get childNodesAsList => JSLiveNodeListWrapper(this, childNodes);
```
srujzs: I don't believe this is a mutable list. Should this be `JSImmutableListWrapper`?

srujzs (commented):
> Does this implementation make sense and matches ideas behind helpers in this module? JSLiveNodeListWrapper don't have to extend JSImmutableListWrapper but have its own implementation but it kinda felt it makes sense this way.

I think so! I'll say that it might better if we started from the `dart:html` implementations instead. I say this mostly because if users are trying to use these helpers to bridge gaps, divergence in behavior would be painful to discover. See `_ChildNodeListLazy` and `_ChildrenElementList`. It's fine if we think some of the `dart:html` implementation methods need to be updated, though.

> could/should NodeList (returned f.e. from querySelector) be extended with asImmutableList via extension? I think it would be more similar to above and dart:html interfaces and JSImmutableListWrapper and JSLiveNodeListWrapper could then be used internally or even marked as private.

Can you elaborate on what you mean here? Do you mean only offer an extension that does the wrapping?

> While doing this I noticed ListMixin and ListBase is going to be deprecated in core? Could implementing List interface directly by both those wrappers while making changes here give any benefits?

I don't think they are but those comments are indeed suspect. @lrhn, is this something we should worry about?

web/lib/src/helpers/lists.dart
fsw: Not sure if this is realiable source:
https://stackoverflow.com/questions/43324751/is-there-a-difference-between-children0-and-firstelementchild
seems difference is only in returned value when list is empty but we throw exceptions in both cases.

I've added `remove`. It was indeed tricky as type of parameter in List interface is `Object?`, please take a look if type checking i used makes sense.

web/lib/src/helpers/lists.dart
```diff
@@ -103,10 +103,22 @@ class JSLiveNodeListWrapper<P extends Node, T extends JSObject, U extends Node>
     parentNode.appendChild(element);
   }
 
+  @override
+  bool remove(Object? element) {
+    // ignore: invalid_runtime_check_with_js_interop_types
```
srujzs: Let's add a todo on me to migrate this to `isJSAny` once we have it.

web/lib/src/helpers/lists.dart
```diff
@@ -69,3 +70,56 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+  void add(U element) {
+    // `ListMixin` implementation only works for lists that allow `null`.
+    parentNode.appendChild(element);
+  }
+
+  @override
+  bool remove(Object? element) {
+    // ignore: invalid_runtime_check_with_js_interop_types
+    if ((element is JSAny?) && (element?.isA<Node>() ?? false)) {
+      if ((element as Node).parentNode == parentNode) {
```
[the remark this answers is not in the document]
srujzs: Try `strictEquals` in `dart:js_interop`, that should do `===` which is more akin to `identical`.

web/lib/src/helpers/lists.dart
srujzs: Ah yes, the list is getting readjusted as you're removing elements. And yeah, I don't think the `end--` makes this any less confusing so this is fine.

web/lib/src/helpers/lists.dart
fsw: done

web/lib/src/helpers/lists.dart
```diff
@@ -65,7 +66,313 @@ class JSImmutableListWrapper<T extends JSObject, U extends JSObject>
...
+  @override
+  _JSList<Element> get _list => _JSList<Element>(_htmlCollection);
+
+  final HTMLCollection _htmlCollection;
+
+  HTMLCollectionListWrapper(this._parent, this._htmlCollection);
+
+  @override
+  Iterator<Element> get iterator =>
+      _HTMLCollectionIterator(_parent.firstElementChild);
```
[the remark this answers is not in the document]
fsw: > Sorry, do you mean that the default iterator implementation checks for concurrent modification but the custom iterators here do not? I think we could fairly easily handle the error-checking in the custom iterators if we really wanted, and it might be worth it to guard users.

Yes but more precisely default list iterator only handles cases where length changes. You can replace elements while iterating as long as length is the same (so it kinda handles some modifications).
With this custom iterators I could add checking of length but this will make the iterator same speed as `ListIterator`. So we would lose the performance benefit we get for not checking it. 

In other words as JS gives us a pointer to the next element (`nextSibling` / `nextElementSibling`) it was tempting to use it for quick iteration but it might have weird unexpected behavior when modifying list that is currently being iterated. For example if you move the `current` element to other parent you will suddenly end up iterating children of this new parent. 

So I guess the question here is if we want faster iterator that is less secure from user errors / unexpected behavior when changing list that is being iterated. Or in other words can we assume programmer should know modifying children when iterating can lead to such issues.

I am not sure myself which option would be better here, I can easily refactor `_filter` to use length + elementAt and remove both those custom iterators for this to be more stable and readable (but bit slower).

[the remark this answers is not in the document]
srujzs: Tricky, especially since we're only guarding against *some* concurrent modificatioins. Even the documentation for `moveNext` in the core library seems to indicate this is optional:

```
  /// A call to `moveNext` may throw for various reasons,
  /// including a concurrent change to an underlying collection.
  /// If that happens, the iterator may be in an inconsistent
  /// state, and any further behavior of the iterator is unspecified,
  /// including the effect of reading [current].
```

So I'm inclined to say it's fine if we don't do the length check, but we should document that on the `iterator` getters.

> In other words as JS gives us a pointer to the next element (nextSibling / nextElementSibling) it was tempting to use it for quick iteration but it might have weird unexpected behavior when modifying list that is currently being iterated.

Is `nextElementSibling`/`nextSibling` faster than `elementAt`? I would assume there's still one interop call that occurs and the latter avoids the issue you described.

I think my preference would be to use `elementAt` if it's not significantly less performant compared to `nextElementSibling`/`nextSibling` and not do the `length` `ConcurrentModificationError` check.

srujzs (commented):
> If we expose all those via extensions should I make NodeListListWrapper, HTMLCollectionListWrapper and JSImmutableListWrapper private? This could make public interface clearer and I do not think any of those would be usable in other way then via this extensions. (in case of JSImmutableListWrapper this will be backward incompatible change so not sure about this one)

We're awaiting a breaking change release so it's fine if we want to hide it, but `JSImmutableListWrapper` can apply to other cases we may be missing (maybe an experimental API for example), so I'm tempted to keep it public. I believe we should make `HTMLCollectionListWrapper` and `NodeListListWrapper` private on reflection though as they're very specific `dart:html` replacements and don't apply to any other member/API. We may even want to add a disclaimer/dart doc explaining that they're meant to be similar to the `dart:html` wrapper behavior on the extension members that return the lists (`childrenAsList` and `childNodesAsList`).

PR #492, opened 2025-12-25
[web_generator] drop the dependency on pkg:crypto and pkg:convert

kevmoo:
Implemented a simple FNV-1a algorithm instead

### DIFF
--- a/web_generator/lib/src/interop_gen/hasher.dart
+++ b/web_generator/lib/src/interop_gen/hasher.dart
@@ -38,8 +38,6 @@ abstract final class AnonymousHasher {
 extension HasherIntExtension on int {
   String to7DigitString() => toString().padLeft(7, '0').substring(0, 7);
 
-  /// Multiplying by `16777619` for the FNV-1a algorithm.
-  ///
   /// 32-bit wraparound multiplication safe for web/JS (53-bit limit).
   int mul32(int multiplier) {
     // Split into 16-bit parts


kevmoo:
> LGTM! Thank you for this!

Gemini did most of the work 😀

web_generator/lib/src/interop_gen/hasher.dart
```diff
@@ -2,67 +2,82 @@
...
+
+    // Combine
+    return (p0 + ((p1 + p2) << 16)) & 0xFFFFFFFF;
+  }
+}
+
+extension on Iterable<String> {
+  /// A hash function based on FNV-1a.
+  ///
+  /// Ensures consistency across platforms vs [Object.hashValues].
```
srujzs: I was going to ask "but why?" but this explains it.

web_generator/lib/src/interop_gen/hasher.dart
```diff
@@ -2,67 +2,82 @@
...
-    radix: 16,
-  ).toInt();
+    // p1 = (aHi * bLo) & 0xFFFF
+    final p1 = (aHi * 403) & 0xFFFF;
+
+    // p2 = (aLo * bHi) & 0xFFFF
+    final p2 = (aLo * 256) & 0xFFFF;
+
+    // Combine
+    return (p0 + ((p1 + p2) << 16)) & 0xFFFFFFFF;
```
srujzs: Shouldn't there be an `aHi * bHi` calculation somewhere in here?

srujzs: Ack, omitting the `aHi` * `bHi` product makes sense due to the modulo. The modification to sum before zeroing bits makes sense to me.

As a nit, @gemini-code-assist, can you change the names of the new variables in your suggestion to be camel-cased?

kevmoo: I'm going to go back to the plain implementation and write some tests.

PR #496, opened 2026-01-04
[generator] Add in a stub to enable wasm compile/run

kevmoo:
It generates great and passes all tests, but it is slower


kevmoo:
Meh. Okay.

PR #498, opened 2026-01-05
[generator] fixes for equality/hashCode

kevmoo:
Remove equal/hash in ast types – several of these were not implemented correctly
And it seems they are not needed. If they ARE needed, then a test should fail

Also did a tiny cleanup one implementation of hashCode

### DIFF
--- a/web_generator/lib/src/ast/types.dart
+++ b/web_generator/lib/src/ast/types.dart
@@ -112,7 +112,6 @@ class TupleType extends ReferredType<TupleDeclaration> {
     name: types.map((t) => t.id.name).join(','),
     index: readonly ? 1 : 0,
   );
-
 }
 
 class UnionType extends DeclarationType {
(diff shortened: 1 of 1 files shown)


kevmoo:
CC @nikeokoronkwo

srujzs:
IIRC, this is relevant in a few spots, but maybe the answer is we just need tests for it.

kevmoo:
> IIRC, this is relevant in a few spots, but maybe the answer is we just need tests for it.

Yeah, because some of the use of `containsAll` etc are not accurate for equality.

nikeokoronkwo:
> IIRC, this is relevant in a few spots, but maybe the answer is we just need tests for it.

The hash codes for the types were relevant because of type deductions (for unions/intersections, for instance), filtering, set operations, and other similar TS features that needed to be mapped to Dart, but most importantly for the former two. 

We should have some more tests (will update), but some of the operations currently have tests for them so if they pass then I think we're okay on those fronts.

PR #499, opened 2026-01-05
Update generated output with current year

kevmoo:
Will make future updates less noisy by doing this in a one-off


srujzs (approved):
https://github.com/dart-lang/web/pull/482 does this as well but you're right that it's cleaner to do it in a separate PR.

Issue #414, opened 2025-07-12
Interop Gen: Support non-identifier properties


nikeokoronkwo:
This issue is to add support for non-identifier properties such as computed property types, string property types, number property types, etc, for the Dart TS Interop Interface Generator:

```ts
interface HttpHeaders {
  "content-type": string;
  [Symbol.iterator]: object;
}
```

nikeokoronkwo:
We can implement string identifier properties.
However, other property kinds will need support on the dart JS interop end: dart-lang/sdk#61413

PR #502, opened 2026-01-07
Generate JS interop bindings for string-literal property names

sagar-h007:
This PR improves JS interop code generation by handling TypeScript interfaces that use string-literal property names
(for example,"content-type" or "Content-Length"). Since these keys can’t be expressed directly as Dart identifiers, the generator now maps them to valid Dart fields using the appropriate `@JS(...)` annotations.

The interop transformer has been updated accordingly, and the existing integration tests have been extended to cover this case and ensure the generated Dart output behaves as expected.

Fixes: #414 

<details>

- I’ve reviewed the Dart contributor guide and made sure this change follows the Dart style guide and existing project conventions.
- This is a small, focused update to interop code generation, so no prior design discussion was needed.
- The code is formatted with `dart format`.
- Existing integration tests were updated to reflect the new behavior.
- No changelog entry or pubspec version update is required for this internal tooling change.

I ran the relevant interop_gen integration tests locally to confirm the generated output matches expectations.

</details>

### DIFF
--- a/web_generator/lib/src/js/typescript.types.dart
+++ b/web_generator/lib/src/js/typescript.types.dart
@@ -534,7 +534,6 @@ extension type TSClassElement._(JSObject _) implements TSNamedDeclaration {
 @JS('PropertyDeclaration')
 extension type TSPropertyDeclaration._(JSObject _)
     implements TSClassElement, TSPropertyEntity {
-  @redeclare
   external TSNode get name;
 }
 
@@ -565,7 +564,6 @@ extension type TSTypeElement._(JSObject _) implements TSNamedDeclaration {
 @JS('PropertySignature')
 extension type TSPropertySignature._(JSObject _)
     implements TSTypeElement, TSPropertyEntity {
-  @redeclare
   external TSNode get name;
   @redeclare
   external TSToken? get questionToken;


web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -2959,3 +2972,16 @@ QualifiedName parseQualifiedName(
     return parseQualifiedNameFromTSQualifiedName(name as TSQualifiedName);
   }
 }
+
+String _toCamelCase(String text) {
```
srujzs: Let's use `dartRename` from `banned_names.dart` to be consistent with IDL generation and also to account for banned keywords. It leads to non-camelcase in some cases, but that's okay.

srujzs (commented):
cc @nikeokoronkwo as well

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -576,14 +576,27 @@ class Transformer {
...
+    // (e.g. `123`) are not supported across many backends effectively.
+    if (nameNode.kind != TSSyntaxKind.Identifier &&
+        nameNode.kind != TSSyntaxKind.StringLiteral) {
+      return null;
+    }
+
     final name = property.name.text;
+    var nameForDart = name;
+    if (nameNode.kind == TSSyntaxKind.StringLiteral) {
+      nameForDart = _toCamelCase(name);
```
nikeokoronkwo: @srujzs is the `@JS` API case sensitive? I don't know if the code would support having the same string but with case-sensitive changes like the following. If it does support it, let's have a test for this, should be able to work for the following example at least:

```ts
interface Headers {
    "content-type": string;
    "Content-type": string;
    "content-Type": string;
    "Content-Type": string;
    "content=type": string;
}
```

nikeokoronkwo (commented):
Thank you @sagar-h007!

srujzs (commented):
LGTM % Nike's comments.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -576,14 +576,27 @@ class Transformer {
     return outputType;
   }
 
-  PropertyDeclaration _transformProperty(
+  PropertyDeclaration? _transformProperty(
     TSPropertyEntity property, {
     required UniqueNamer parentNamer,
     TypeDeclaration? parent,
   }) {
+    final nameNode = property.name as TSNode;
```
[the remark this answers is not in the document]
sagar-h007: I’ve updated the type in typescript.types.dart to avoid the cast and model it as a proper union, as suggested. Thanks!

nikeokoronkwo (approved):
LGTM, Thanks @sagar-h007

srujzs (approved):

Issue #465, opened 2025-09-09
Interop Gen: Support Complex Literal Type Expressions


nikeokoronkwo:
This issue is for adding support to the TS Declaration Generator for additional literal type kinds such as:
- [x] prefix unary expressions: `-1`
- [ ] binary expressions: `1+2`
- [ ] others

This is meant for expressions containing a combination of these expressions, particularly

Varadraj75:
Thanks for the clarification, @sagar-h007.

From testing on main, unary literal cases (e.g. -1) already work via the type checker and existing tests pass.

If this issue is meant to track support for computed/binary literal expressions (like 1 + 2, 1 << 16), could we clarify that explicitly? Otherwise, it seems the original scope may already be covered.

@nikeokoronkwo — please confirm whether this should remain open for computed literals or be closed and tracked separately.

nikeokoronkwo:
> Hey [@nikeokoronkwo](https://github.com/nikeokoronkwo) , I tested this on the current main branch and the existing literal_test passes (including unary numeric literals like -1 resolved via the type checker). Is there a specific remaining case or scenario that is still not handled and that this issue is intended to track?

The case for computed literals is not completed: not many changes were done concerning the addition of features to the generator in #503 and so it was more of a refactor.

PR #503, opened 2026-01-11
Interop gen additional literal types

sagar-h007:
This PR improves interop_gen support for TypeScript literal types that aren’t represented as simple literals in the AST (for example, prefix unary expressions like -1).

When a literal can’t be handled directly, the generator now falls back to the TypeScript TypeChecker to resolve the type. If the resolved type is a literal, its value is used; otherwise, the code safely falls back to the underlying Dart type, preserving existing behavior.

The change also adds the minimal TypeScript bindings needed for this resolution and includes an integration test to cover the new case and prevent regressions.

Fixes:#465

---

- I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

	•	This is a small, targeted change based on an existing issue.
	•	The code follows the Dart style guide and is formatted with dart format.
	•	New integration tests were added, and all existing tests pass locally.
	•	No changelog or pubspec update is needed for this internal tooling change.

</details>

### DIFF
--- a/web_generator/test/integration/interop_gen/literal_test_input.d.ts
+++ b/web_generator/test/integration/interop_gen/literal_test_input.d.ts
@@ -3,7 +3,7 @@ export declare const negativeOne: -1;
 // export declare const binaryAdd: 1 + 2;
 // export declare const shiftLeft: 1 << 16;
 // export declare const complex: 1 + 2 * 3;
-// export declare const existingLiteral: 1;
+export declare const existingLiteral: 1n;
 export declare const existingString: "hello";
 export declare const existingBoolean: true;
 export declare const falsy: false;
--- a/web_generator/test/integration/interop_gen/literal_test_expected.dart
+++ b/web_generator/test/integration/interop_gen/literal_test_expected.dart
@@ -6,6 +6,8 @@ import 'dart:js_interop' as _i1;
 @_i1.JS()
 external num get negativeOne;
 @_i1.JS()
+external _i1.JSAny? get existingLiteral;
+@_i1.JS()
 external String get existingString;
 @_i1.JS()
 external bool get existingBoolean;


web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -1742,34 +1742,96 @@ class Transformer {
...
+                    kind: LiteralKind.$true,
+                    value: true,
+                  );
+                } else if (typeStr == 'false') {
+                  return LiteralType(
+                    isNullable: isNullable ?? false,
+                    kind: LiteralKind.$false,
+                    value: false,
+                  );
+                }
```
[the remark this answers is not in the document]
sagar-h007: I’ll simplify this block to reduce duplication.

web_generator/test/integration/interop_gen/literal_test_input.d.ts
```diff
@@ -0,0 +1,8 @@
+export declare const negativeOne: -1;
+// Binary expressions in type position are not valid TS in d.ts files usually.
```
srujzs: Do we have an example we can test where it *is* valid?

web_generator/lib/src/js/typescript.types.dart
```diff
@@ -15,6 +15,69 @@ import 'annotations.dart';
 import 'helpers.dart';
 import 'typescript.dart';
 
+extension type const TSTypeFlags._(int _) implements int {
```
srujzs: nit: I think I prefer omitting unused constants with maybe a TODO to add more.

web_generator/test/integration/interop_gen/literal_test_input.d.ts
```diff
@@ -0,0 +1,8 @@
...
+// export declare const binaryAdd: 1 + 2;
+// export declare const shiftLeft: 1 << 16;
+// export declare const complex: 1 + 2 * 3;
+export declare const existingLiteral: 1;
+export declare const existingString: "hello";
+export declare const existingBoolean: true;
```
srujzs: Let's add a test for `false` and `null` as well.

web_generator/lib/src/interop_gen/transform/transformer.dart
```diff
@@ -1742,34 +1742,90 @@ class Transformer {
...
+                  return LiteralType(
+                    isNullable: isNullable ?? false,
+                    kind: value ? LiteralKind.$true : LiteralKind.$false,
+                    value: value,
+                  );
+                }
+              }
+
+              // Fallback to underlying type if not a literal
+              final underlyingTypeNode = typeChecker.typeToTypeNode(
```
srujzs: Is there a test exercising this case?

sagar-h007: Not yet ,. I’ll add a test to explicitly exercise this fallback path.

web_generator/lib/src/js/typescript.types.dart
sagar-h007: I’ll drop the unused constants for now and leave a TODO for adding more when we actually need them. That should keep things simpler and safer.

srujzs (commented):
Mostly LGTM % my comments.

web_generator/lib/src/interop_gen/transform/transformer.dart
srujzs: I might have missed it, but which test case covers this?

srujzs:
Ack, I think the only thing left is the uncomment (and generate the expectation file) and we can go ahead and land this.

sagar-h007:
@srujzs Done — re-enabled the test using a 1n BigInt to hit the fallback path and updated the expected output. All tests are passing.

PR #522, opened 2026-03-15
[generator] Add support for source map backed stack traces for debugging generator output

### LINKED ISSUE #505: [web_generator] Improve debugging of bin scripts with better stack traces
The stack traces produced by JS code are usually difficult to map/trace to Dart, and since we plan on maybe making this accessible to users as a public package (see #475), we shouldn't place this burden on users, and instead we should make it possible for easy debugging of this library.

We can use some of the JS types (`error.stack` if available else use [`Error.captureStackTrace()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/captureStackTrace) to add the trace) to get the stack trace in JS, parse the trace (either ourselves, or there is a popular [`stack-trace`](https://www.npmjs.com/package/stack-trace) library we can use), and then use the `source_map_stack_trace` package to take the trace and the JS source map to produce a Dart-compatible source trace.

> Although stack traces are [not yet standardized](https://github.com/tc39/proposal-error-stacks) in JS, they are widely implemented, and that includes V8 (`node`).

Although these stack traces are more helpful, they are not perfect and accurate, and while I'll try to improve the traces presented as best as possible, until they show a completely accurate representation of the code, we should add a `--raw-traces` (for lack of better name; name is tentative) flag to allow users to print the raw trace when reporting issues or bugs (or if we/they want to try to debug), which falls back to this normal trace.

@srujzs Wanted to know if this is something that would make sense before working on it.

nikeokoronkwo:
Fixes #505

This PR adds support for working with the web generator errors better using dart-style stack traces by mapping the stack trace frames from V8/NodeJS to Dart, using the related dart packages.

This makes the error messages much easier to debug and deal with, which improves debugging

## Issues
1. There are some issues mapping the frames completely, including but not limited to mapping `npm` dependency code in the stack frames. The major example here being that the frames occuring from the `typescript` package are being mapped to other unrelated packages. An example before:
  ```
      at getTypeFromTypeReference (~/web/web_generator/lib/src/node_modules/typescript/lib/typescript.js:65071:36)
      at getTypeFromTypeNodeWorker (~/web/web_generator/lib/src/node_modules/typescript/lib/typescript.js:67750:16)
      at getTypeFromTypeNode (~/web/web_generator/lib/src/node_modules/typescript/lib/typescript.js:67712:41)
      at Object.getTypeFromTypeNode (~/web/web_generator/lib/src/node_modules/typescript/lib/typescript.js:51140:21)
      at Transformer._transformType$4$isNullable$parameter$typeArg (~/web/web_generator/lib/src/dart_main.js:8064
      at Transformer._transformType$1 (~/web/web_generator/lib/src/dart_main.js:81095:19)
  ```
  
  and after:
  ```
  ../../../../../../.pub-cache/hosted/pub.dev/dart_style-3.1.3/lib/src/front_end/piece_writer.dart 208:23  getTypeFromTypeReference
  ../../../../../../.pub-cache/hosted/pub.dev/dart_style-3.1.3/lib/src/short/rule/rule.dart 166:5          getTypeFromTypeNode
  org-dartlang-sdk:///lib/_internal/js_shared/lib/js_util_patch.dart 149:10                                Transformer._transformType
  ```


kevmoo:
@nikeokoronkwo – you going to run w/ this?

nikeokoronkwo:
I will be back to this soon. I need to report an issue to the source_map_stack_traces library first, and I asked Srujan if he could take a look before I moved this from a draft. I have not gotten a response yet from him.

PR #526, opened 2026-04-06
Update min SDKs, regenerate

kevmoo:
Move pkg:web to Dart 3.10 and regenerate
Move pkg:js_interop to Dart 3.10
Setup to only compile and run `npm install` at most once per test run.

### DIFF
--- a/web_generator/test/sdk_version_test.dart
+++ b/web_generator/test/sdk_version_test.dart
@@ -6,20 +6,26 @@
 library;
 
 import 'dart:io';
+import 'dart:isolate';
+import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:test/test.dart';
 import 'package:web_generator/src/sdk_version.dart';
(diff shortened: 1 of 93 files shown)


web_generator/test/sdk_version_test.dart
```diff
@@ -0,0 +1,71 @@
...
+      _ => throw const FormatException(
+        'No sdk constraint found in pubspec.yaml',
+      ),
+    };
+
+    final sdkConstraint = VersionConstraint.parse(sdkConstraintStr);
+
+    // Check that sdkVersion satisfies the constraint
+    expect(
+      sdkConstraint.allows(sdkVersion),
```
srujzs: I may be unfamiliar with this API, but shouldn't we not need this test with the test below?

srujzs (approved):
Nice, thanks for the cleanup!

web_generator/test/sdk_version_test.dart
kevmoo: good catch! yep!

PR #529, opened 2026-04-14
Iterable and Map support

kevmoo:


kevmoo:
I have better things coming!

PR #530, opened 2026-04-14
generator: add some sanity to the language version logic

kevmoo:


web_generator/test/sdk_version_test.dart
```diff
@@ -37,26 +37,28 @@ void main() {
...
       reason:
-          'sdkVersion $sdkVersion does not match expected version '
-          '$expectedVersion derived from minimum of constraint '
+          'dartLangugeVersion $dartLanguageVersion does not match expected '
+          'version $expectedVersion derived from minimum of constraint '
           '$sdkConstraint',
     );
   });
+
+  test('language version derived from SDK constraint', () {
```
srujzs, line 52: Don't know how much it's worth it, but you can create a fake `pubspec.yaml` and then test `checkSdkVersion` directly. Then, you don't need to test `deriveLanguageVersion` and can just make it private. You can also test that `SdkVersionException`s are thrown that way.

srujzs (approved):

web_generator/test/sdk_version_test.dart
kevmoo: Wait...explain this? Just for testing?

web_generator/lib/src/sdk_version.dart
```diff
@@ -2,7 +2,64 @@
...
+
+/// Derives the language version from an SDK constraint.
+Version deriveLanguageVersion(VersionConstraint constraint) {
+  final minVersion = switch (constraint) {
+    Version() => constraint,
+    VersionRange(min: final min) => min ?? Version(0, 0, 0),
+    _ => throw ArgumentError(
+      'Unsupported constraint type: ${constraint.runtimeType}',
+    ),
+  };
```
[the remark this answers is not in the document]
kevmoo: Don't care about unions!

web_generator/test/sdk_version_test.dart
kevmoo: Happy to do a follow-up!

srujzs: Yup, just so we're testing the feature actually being used in the generators instead of a subfunction (`checkSdkVersion` vs `deriveLanguageVersion`).

PR #531, opened 2026-04-14
Rename web_generator -> interop_generator

kevmoo:


kevmoo:
I'm going to come BACK with `web_generator` just for pkg:web

PR #534, opened 2026-04-19
feat: generate descriptive documentation for union typedefs

### LINKED ISSUE #337: docs request: help me figure out what to do with JS unions when they occur
As an example, [the docs for `ReadableStream.getReader()`](https://pub.dev/documentation/web/latest/web/ReadableStream/getReader.html) says that it returns a `ReadableStreamReader`. Cool.

Except [the docs for `ReadableStreamReader`](https://pub.dev/documentation/web/latest/web/ReadableStreamReader.html) says that it's just a `JSObject`. 

Now what am I supposed to do?

After talking to @kevmoo, I learned that this happens when there's a JS union, which Dart doesn't support. Fair enough. I also learned that [the MDN docs for `getReader()`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/getReader#return_value) tell me that the possible JS returns are `ReadableStreamDefaultReader` or `ReadableStreamBYOBReader`, both of which have Dart equivalents. However *neither* of these types is mentioned in the Dart web `getReader()` docs or the `ReadableStreamReader` docs, so how do I know that if I don't talk to Kevin?

With this info, my first thought is to do a Dart cast to `ReadableStreamDefaultReader` (the default) but according to Kevin, I'm not supposed to do that either. Apparently [there's some lint that tells me what to do](https://dart.dev/tools/linter-rules/invalid_runtime_check_with_js_interop_types). Which is great, if I've got that lint enabled.

But even if I do have that lint enabled, the docs should tell me not to waste my time on Dart `as` or `is` to learn from the lint (if I have it enabled) what the right thing to do is.

The MDN docs tell me what the right thing to do is. And it's my understanding that the Dart docs are generated from the MDN docs (or the same metadata that the MDN docs are generated from). **So can we please make the Dart docs tell me what the right thing to do in the face of a JS union is?**

Thank you. And thank you for attending my Ted talk.

kevmoo:
- Added generator logic to document the types that make up a union typedef.
- Mapped Web IDL primitives and special types to their corresponding `dart:js_interop` types (e.g., DOMString -> JSString, Function -> JSFunction) in docs.
- Wrapped types not generated in the package in backticks to avoid broken links.
- Correctly placed @docImport directives on the library directive using b.docs.

Fixes https://github.com/dart-lang/web/issues/337

### DIFF
--- a/js_interop_gen/lib/src/translator.dart
+++ b/js_interop_gen/lib/src/translator.dart
@@ -19,16 +19,6 @@ import 'type_aliases.dart';
 import 'type_union.dart';
 import 'util.dart';
 
-final Map<String, Set<String>> _superToSubtypes = () {
-  final map = <String, Set<String>>{};
-  jsTypeSupertypes.forEach((subtype, supertype) {
-    if (supertype != null) {
-      map.putIfAbsent(supertype, () => {}).add(subtype);
-    }
-  });
-  return map;
-}();
-
 typedef TranslationResult = Map<String, code.Library>;
 
 class _Library {
@@ -761,8 +751,6 @@ class Translator {
   /// Singleton so that various helper methods can access info about the AST.
   static Translator? instance;
 
-  static const _unionDocListThreshold = 4;
-
   Translator(
     this._librarySubDir,
     this._cssStyleDeclarations,
@@ -941,16 +929,25 @@ class Translator {
     }
 
     final typeNames = types.map(_getTypeNameRaw).toList();
-    final collapsedNames = _collapseTypes(typeNames);
+    final uniqueNames = typeNames.toSet().toList();
 
-    final formattedNames = collapsedNames.map((name) {
+    if (uniqueNames.length <= 1) return [];
+
+    final formattedNames = uniqueNames.map((name) {
       final decl = _typeToDeclaration[name];
       if (decl != null && _usedTypes.contains(decl)) {
         return '[$name]';
       }
-      // If it's a generic type (contains <), wrap in ticks to avoid HTML.
+      // If it's a generic type (contains <), use fancy formatting.
       if (name.contains('<')) {
-        return '`$name`';
+        final parts = name.split('<');
+        final base = parts[0];
+        final generic = parts[1].replaceAll('>', '');
+        final genericParts = generic.split(',').map((s) => s.trim());
+        final linkedGenericParts = genericParts
+            .map((part) => '[$part]')
+            .join(', ');
+        return '<code>[$base]\\<$linkedGenericParts\\></code>';
       }
       // Link if it's a mapped primitive or a valid JS interop type from
       // supertypes map.
@@ -961,16 +958,10 @@ class Translator {
       return '`$name`';
     }).toList();
 
-    // Sort: alphabetical, but ticked ones last!
-    formattedNames.sort((a, b) {
-      final aIsTicked = a.startsWith('`');
-      final bIsTicked = b.startsWith('`');
-      if (aIsTicked && !bIsTicked) return 1;
-      if (!aIsTicked && bIsTicked) return -1;
-      return a.compareTo(b);
-    });
+    formattedNames.sort();
 
-    if (formattedNames.length >= _unionDocListThreshold) {
+    final singleLine = '/// Union of: ${formattedNames.join(', ')}';
+    if (singleLine.length > 80) {
       return [
         '/// Union of ${formattedNames.length} types',
         '///',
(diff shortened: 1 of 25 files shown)


js_interop_gen/test/integration/idl/union_docs_expected.dart
```diff
@@ -0,0 +1,35 @@
+// Generated from Web IDL definitions.
+
+// ignore_for_file: constant_identifier_names, non_constant_identifier_names
+
+@JS()
+library;
+
+import 'dart:js_interop';
+
+/// Union of: [JSTypedArray]
```
srujzs: I think we should omit the docs if the resulting type is just one type.

js_interop_gen/lib/src/translator.dart
```diff
@@ -918,12 +932,166 @@ class Translator {
...
+        '/// Union of ${formattedNames.length} types',
+        '///',
+        for (final name in formattedNames) '/// - $name',
+      ];
+    }
+
+    return ['/// Union of: ${formattedNames.join(', ')}'];
+  }
+
+  List<String> _collapseTypes(List<String> types) {
```
srujzs: Feel free to push back, but I think this might give users the wrong idea that the supertypes are complete. For example, we don't actually provide every subtype of `TypedArray` that JS provides through `dart:js_interop`. If the documentation claims it's only the ones we provide, then we should document each of those. Therefore, I think we can remove this and just remove duplicates instead.

js_interop_gen/lib/src/translator.dart
```diff
@@ -918,12 +932,166 @@ class Translator {
...
+      // Link if it's a mapped primitive or a valid JS interop type from
+      // supertypes map.
+      if (_mapIdlPrimitiveToDart(name) != null ||
+          jsTypeSupertypes.containsKey(name)) {
+        return '[$name]';
+      }
+      return '`$name`';
+    }).toList();
+
+    // Sort: alphabetical, but ticked ones last!
```
srujzs: Sort is likely good to reduce noise, but why ticked ones last? :)

js_interop_gen/lib/src/translator.dart
```diff
@@ -918,12 +932,166 @@ class Translator {
...
+    formattedNames.sort((a, b) {
+      final aIsTicked = a.startsWith('`');
+      final bIsTicked = b.startsWith('`');
+      if (aIsTicked && !bIsTicked) return 1;
+      if (!aIsTicked && bIsTicked) return -1;
+      return a.compareTo(b);
+    });
+
+    if (formattedNames.length >= _unionDocListThreshold) {
```
srujzs: Let's maybe do a character limit instead so that we stay within 80 chars. All typedefs will be top-level so just checking that the string length will be <= than 80 chars is fine.

js_interop_gen/lib/src/translator.dart
```diff
@@ -918,12 +932,166 @@ class Translator {
...
+      return;
+    }
+    final name = (idlType.idlType as JSString).toDart;
+    final library = _typeToLibrary[name];
+    if (library != null && library.url != _currentlyTranslatingUrl) {
+      _currentDocImports.add(library.url);
+    }
+  }
+
+  String? _mapIdlPrimitiveToDart(String idlType) {
```
srujzs, line 990: Maybe repurpose https://github.com/dart-lang/web/blob/db91caf29bfd00c509179a6298905eb84e177924/js_interop_gen/lib/src/type_aliases.dart#L5 somehow instead?

js_interop_gen/lib/src/translator.dart
```diff
@@ -1544,9 +1712,52 @@ class Translator {
...
+
     if (_generateForWeb) {
       b.comments.addAll([...licenseHeader, '', ...mozLicenseHeader]);
     }
+
+    if (docImports.isNotEmpty) {
+      final currentDir = p.url.dirname(_currentlyTranslatingUrl);
+      b.docs.addAll([
+        for (final url in docImports)
+          '/// @docImport \'${p.url.relative(url, from: currentDir)}\';',
```
srujzs, line 1703: Cool! I didn't know about this.

js_interop_gen/lib/src/translator.dart
```diff
@@ -918,12 +932,166 @@ class Translator {
...
+    final collapsedNames = _collapseTypes(typeNames);
+
+    final formattedNames = collapsedNames.map((name) {
+      final decl = _typeToDeclaration[name];
+      if (decl != null && _usedTypes.contains(decl)) {
+        return '[$name]';
+      }
+      // If it's a generic type (contains <), wrap in ticks to avoid HTML.
+      if (name.contains('<')) {
+        return '`$name`';
```
srujzs: If you want to be really fancy, you can do something like:

```
<code>[JSArray]\<JSObject\></code>
```

srujzs (approved):
Generally LGTM % my comments. Thanks Kevin!

js_interop_gen/test/integration/idl/union_docs_expected.dart
kevmoo: 😆 I didn't catch that one! I agree!!!

js_interop_gen/lib/src/translator.dart
kevmoo: Because they are not available in our APIs. So we put them in a place of shame.

js_interop_gen/lib/src/translator.dart
kevmoo: Thank @srawlins !

js_interop_gen/lib/src/translator.dart
kevmoo: I made it even better!

js_interop_gen/lib/src/translator.dart
kevmoo: PTAL

js_interop_gen/lib/src/translator.dart
kevmoo: I fixed.

js_interop_gen/lib/src/translator.dart
kevmoo: look again?

PR #536, opened 2026-04-19
refactor(js_interop_gen): big translator refactor and cleanups

kevmoo:
- Moved the special-case mapping of `WindowProxy` to `Window` from hardcoded checks in `translator.dart` to the centralized `idlOrBuiltinToJsTypeAliases` map in `type_aliases.dart`.
- Decoupled type name resolution and documentation generation logic from `translator.dart` into separate files (like `type_mapper.dart`) to improve maintainability.
- Added new integration tests for inheritance fallback and name collisions in Web IDL.
- Improved doc import tracking to automatically omit `/// @docImport` lines for libraries that are already included via regular imports.
- Regenerated bindings in `pkg:web`, resulting in cleaner output with less redundant doc imports across multiple files.


nshahan:
Meta level drive by nit: 

PR descriptions are helpful when reviewing. What is the nature of the big refactor?

Especially if/when this land it would be nice to have helpful descriptions to refer to in commit history.

PR #540, opened 2026-04-28
js_interop_gen: minimize the number of `@docImport` comments we create

kevmoo:
No need to create these if there is already a library import

Confirmed that we still have the same 811 doc issues before/after the change
So this makes things no worse!

### DIFF
--- a/js_interop_gen/lib/src/translator.dart
+++ b/js_interop_gen/lib/src/translator.dart
@@ -1120,10 +1120,7 @@ class Translator {
       );
     }
     final url = _urlForType(dartType);
-    final originalUrl = _typeToLibrary[dartType]?.url;
-    if (originalUrl != null) {
-      _regularImports.add(originalUrl);
-    }
+
     return code.TypeReference(
       (b) => b
         ..symbol = dartType
@@ -1139,6 +1136,9 @@ class Translator {
     // Unfortunately, `code_builder` doesn't know the url of the library we are
     // emitting, so we have to remove it here to avoid importing ourselves.
     var url = _typeToLibrary[dartType]?.url;
+    if (url != null) {
+      _regularImports.add(url);
+    }
 
     // JS types and core types don't have urls.
     if (url == null) {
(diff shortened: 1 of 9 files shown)


web/analysis_options.yaml
```diff
@@ -12,7 +12,7 @@ analyzer:
   errors:
     # 43 instances in generated code.
     camel_case_types: ignore
-    # 420 instances in the MDN docs.
+    # 811 instances in the MDN docs.
     comment_references: ignore
```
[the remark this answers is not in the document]
kevmoo: We want it for tracking. Leaving it!

js_interop_gen/lib/src/translator.dart
```diff
@@ -739,6 +739,7 @@ class Translator {
   final _usedTypes = <idl.Node>{};
   final _renamedClasses = <String, String>{};
   final _currentDocImports = <String>{};
+  final _regularImports = <String>{};
```
srujzs: I can't think of a good way to pull _currentDocImports and _regularImports out in a way that is local to the `Library`, so maybe just rename `_regularImports` to `_currentLibraryImports` so it's obvious this gets reinitialized.

js_interop_gen/lib/src/translator.dart
```diff
@@ -1134,6 +1135,9 @@ class Translator {
     // Unfortunately, `code_builder` doesn't know the url of the library we are
     // emitting, so we have to remove it here to avoid importing ourselves.
     var url = _typeToLibrary[dartType]?.url;
+    if (url != null) {
```
srujzs: nit: Let's just put this above line 1151 where we already check if it's non-null and not the current url. Including the current url in something called "imports" is a bit awkward.

js_interop_gen/lib/src/translator.dart
```diff
@@ -1687,10 +1692,11 @@ class Translator {
         ..._interfacelike(interfacelike),
     ];
 
-    final docImports = <String>[];
-    if (_currentDocImports.isNotEmpty) {
-      docImports.addAll(_currentDocImports.toList()..sort());
-    }
+    final docImports =
+        _currentDocImports
+            .where((url) => !_regularImports.contains(url))
```
srujzs: If we move the above check as described, this will also need to check if the url is `library.url`.

srujzs (approved):

PR #541, opened 2026-04-28
test(js_interop_gen): add integration tests

kevmoo:
- Added integration tests for inheritance fallback and name collisions in Web IDL:
  - `inheritance_fallback_input.idl` tests inheriting from a super type not defined in the IDL.
  - `name_collision_input.idl` tests operations with the same name as a defined type.
- Marked one of them as skipped because it reveals a crash. Fixing the crash messes up the existing handling of `Sensor`. So it's a reminder to revisit after the refactor.

### DIFF
--- a/js_interop_gen/test/integration/idl_test.dart
+++ b/js_interop_gen/test/integration/idl_test.dart
@@ -63,9 +63,8 @@ void main() {
             from: bindingsGenPath,
           );
 
-          // TODO(kevmoo): Do a more complete cleanup to remove dependency on BCD
-          // and webref bits entirely from js_interop_gen tests.
-          // run the entrypoint
+          // TODO(kevmoo): Do a more complete cleanup to remove dependency on
+          // BCD and webref bits entirely from js_interop_gen tests.
           final dummyBcdPath = p.relative(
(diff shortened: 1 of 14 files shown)


js_interop_gen/test/integration/idl_test.dart
```diff
@@ -45,32 +45,46 @@ void main() {
...
+            from: bindingsGenPath,
+          );
+          final outFilePath = p.relative(
+            outputActualPath,
+            from: bindingsGenPath,
+          );
+
+          // TODO(kevmoo): Do a more complete cleanup to remove dependency on
+          // BCD and webref bits entirely from js_interop_gen tests.
+          final dummyBcdPath = p.relative(
```
srujzs, line 68: IIRC, this no longer exists, so should we get rid of this argument to `main.mjs` and let the arg handling handle this?

srujzs (approved):

js_interop_gen/test/integration/idl_test.dart
kevmoo: Will look! (although this line change is unrelated, it just changed the format)

Let me follow-up in another PR!

PR #542, opened 2026-04-28
js_interop_gen: Move WindowProxy logic to the existing

kevmoo:
...type aliases map.

It's cleaner

### DIFF
--- a/js_interop_gen/lib/src/translator.dart
+++ b/js_interop_gen/lib/src/translator.dart
@@ -744,7 +744,7 @@ class Translator {
   final _usedTypes = <idl.Node>{};
   final _renamedClasses = <String, String>{};
   final _currentDocImports = <String>{};
-  final _regularImports = <String>{};
+  final _currentLibraryImports = <String>{};
 
   Map<String, String> get renamedClasses => _renamedClasses;
 
@@ -1142,7 +1142,7 @@ class Translator {
     // emitting, so we have to remove it here to avoid importing ourselves.
     var url = _typeToLibrary[dartType]?.url;
     if (url != null) {
-      _regularImports.add(url);
+      _currentLibraryImports.add(url);
     }
 
     // JS types and core types don't have urls.
@@ -1154,6 +1154,7 @@ class Translator {
     } else if (url == _currentlyTranslatingUrl) {
       url = null;
     } else {
+      _currentLibraryImports.add(url);
       url = p.url.relative(url, from: p.url.dirname(_currentlyTranslatingUrl));
     }
     return url;
@@ -1669,7 +1670,7 @@ class Translator {
 
   code.Library _library(_Library library) => code.Library((b) {
     _currentDocImports.clear();
-    _regularImports.clear();
+    _currentLibraryImports.clear();
 
     final body = [
       for (final typedef in library.typedefs.where(_usedTypes.contains))
@@ -1700,7 +1701,7 @@ class Translator {
 
     final docImports =
         _currentDocImports
-            .where((url) => !_regularImports.contains(url))
+            .where((url) => !_currentLibraryImports.contains(url))
             .toList()
           ..sort();
 
--- a/js_interop_gen/lib/src/translator.dart
+++ b/js_interop_gen/lib/src/translator.dart
@@ -1120,10 +1120,7 @@ class Translator {
       );
     }
     final url = _urlForType(dartType);
-    final originalUrl = _typeToLibrary[dartType]?.url;
-    if (originalUrl != null) {
-      _regularImports.add(originalUrl);
-    }
+
     return code.TypeReference(
       (b) => b
         ..symbol = dartType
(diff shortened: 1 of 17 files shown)


srujzs (approved):

js_interop_gen/lib/src/translator.dart
```diff
@@ -229,9 +229,6 @@ _RawType _getRawType(idl.IDLType idlType) {
   // Handles types that don't exist in the set of IDL type declarations. They
   // are either some special values or JS builtin types.
 
-  // `WindowProxy` doesn't exist as an interface in the IDL. For our purposes,
-  // `Window` is the appropriate interface.
-  if (type == 'WindowProxy') type = 'Window';
   // `any` is marked non-nullable in the IDL, but since it is a union of
   // `undefined`, it can be nullable for our purposes.
   if (type == 'any') nullable = true;
```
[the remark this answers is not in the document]
kevmoo: Good catch!

kevmoo:
@srujzs – you might want to look again, given the agent comment and the follow-up fix.

js_interop_gen/lib/src/translator.dart
```diff
@@ -229,13 +229,18 @@ _RawType _getRawType(idl.IDLType idlType) {
...
   final translator = Translator.instance!;
+
+  // Resolve aliases that point to other IDL types so they can be marked as used
```
srujzs: I think this is right but better written as:

```dart
  final alias = idlOrBuiltinToJsTypeAliases[type];
  if (alias != null) {
    // The declaration should exist if there is no alias.
    assert(translator._typeToDeclaration[type] != null);
  } else {
    type = alias;
  }
  if (!translator.markTypeAsUsed(type)) {
    // If the type is an IDL type that is never generated, use its JS type
    // equivalent.
    type = _getJSTypeEquivalent(_RawType(type, false))!.type;
  }
  return _RawType(type, nullable, typeParameter);
```

js_interop_gen/lib/src/translator.dart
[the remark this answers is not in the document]
srujzs: > If Window is only referenced via WindowProxy in the IDL, the Window interface itself may not be generated in the output.

I think this will functionally never happen but it's a good note. I added a block of code that I think would be better.

js_interop_gen/lib/src/translator.dart
kevmoo: HA! This is what the robot said:

***

Thanks for the suggestion! I took a close look at this refactor, and I think we should stick with the current implementation for a couple of reasons:

1. **Typo in the suggestion**: In the `if-else` block, the `else` branch sets `type = alias` when `alias` is actually `null`:
   ```dart
   if (alias != null) {
     // ...
   } else {
     type = alias; // This makes `type` null!
   }
   ```
   We would need to fix that to avoid setting `type` to null.

2. **Logic Flaw / Regression Risk**: Even if we fix the typo, the suggestion changes the behavior for types that alias to built-in JS types rather than other IDL declarations.
   The current code only updates `type` to the alias if the alias itself has a declaration in the IDL (like `WindowProxy` aliasing to `Window`):
   ```dart
   final aliasCheck = idlOrBuiltinToJsTypeAliases[type];
   if (aliasCheck != null &&
       translator._typeToDeclaration.containsKey(aliasCheck)) {
     type = aliasCheck;
   }
   ```
   If we remove that check and unconditionally set `type = alias` for something like `sequence` (which aliases to `JSArray`), then `type` becomes `JSArray`.
   Later in the method, `markTypeAsUsed('JSArray')` will return `false` (since it's not in `_typeToDeclaration`). Then it will call `_getJSTypeEquivalent('JSArray')`, which returns `null` for non-declarations, leading to a crash on the null-check operator (`!`) on this line:
   ```dart
   type = _getJSTypeEquivalent(_RawType(type, false))!.type;
   ```

Given that, the current code is safer as it prevents this fall-through crash for built-in JS types.

srujzs: 1. Yes, this is a typo. It should check `== null`.
2. Ah, that's why we had the `alias == null` check. That can be added back. We could avoid the `!` and check the result, but the following is cleaner.

@gemini-code-assist, what are your thoughts on this revision?

```dart
  final alias = idlOrBuiltinToJsTypeAliases[type];
  if (alias == null) {
    // The declaration should exist if there is no alias.
    assert(translator._typeToDeclaration[type] != null);
    if (!translator.markTypeAsUsed(type)) {
      // If the type is an IDL type that is never generated, use its JS type
      // equivalent.
      type = _getJSTypeEquivalent(_RawType(type, false))!.type;
    }
  } else {
    type = alias;
  }
  return _RawType(type, nullable, typeParameter);
```

srujzs: Actually, this doesn't work quite right. We don't mark `Window` as used now because it's an alias.

Looking at `idlOrBuiltinToJsTypeAliases` again, all the values are JS types. I don't think that's where we want `Window` anyways. We can have another map, but that seems like the wrong map to put it in.

kevmoo: ha! so you were WRONG??

Should I revert one then? Or just revert this WHOLE THING?

srujzs: I do like my slight refactor because the current code is a little confusing, but we should still keep the `if (type == 'WindowProxy') type = 'Window';` line. Outside of the refactor (and up to you if you want to even bother landing that), yeah this PR should be reverted.

PR #549, opened 2026-05-13
js_interop_gen: many fixes moving towards supporting large, complex D.TS files

kevmoo:
- **fix: resolve namespace flattening type resolution mismatch in js_interop_gen**
- **fix(js_interop_gen): traverse generic type arguments in ReferredType dependency resolver**
- **refactor(js_interop_gen): eliminate GlobalOptions mutable state in favor of options propagation**
- **refactor(js_interop_gen): decompose transformer.dart by extracting DependencyWalker**
- **refactor(js_interop_gen): decompose transformer.dart by extracting NamespaceFlattener**
- **fix(js_interop_gen): prevent self-import and prefixing when importUrl matches file**
- **fix(js_interop_gen): deduplicate collected generic parameters by name in unions**
- **fix(js_interop_gen): distinguish generic builtins by type arguments and cast distinct union array getters**
- **fix(js_interop_gen): map void generic type arguments to JSAny to conform to JSAny? bounds**
- **feat(js_interop_gen): support string-literal module declarations**
- **feat(js_interop_gen): prevent infinite recursion crashes on mutually recursive types**
- **feat(js_interop_gen): support generic type parameter resolution inside module namespaces**
- **feat(js_interop_gen): resolve enum member type references to parent EnumDeclaration**
- **feat(js_interop_gen): add formatter try-catch recovery and document next steps**
- **feat(js_interop_gen): resolve literal union casts and refactor AST Node type safety**
- **chore: move type resolver into its own file**
- **feat: cache transformed AST nodes to resolve mismatched unique-suffix references**
- **fix: deduplicate types in union/intersection declarations to resolve duplicate definition errors**
- **fix: resolve unchecked use of nullable value errors for JSAny and void in union getters**
- **fix: uniquely identify generic type references in dependency walking to prevent missing generic unions**
- **fix: pad generic type arguments in ReferredType emit to match declaration bounds count**
- **fix: prevent parent nullability leaking into generic type arguments during referred type emit**
- **fix: always emit Deprecated annotation with a default message argument**
- **fix: dynamically resolve and enforce non-JSObject representation types for interfaces**
- **fix: resolve all static analysis lints and line-length limits in lib/**
- **fix: automatically detect and redeclare conflicting inherited members inside intersection types**
- **fix: cast generic ReferredTypes using intermediate JSAny cast in union getters**
- **test: replace global Object inheritance with custom Parent interface in representation type test to avoid TS warning**
- **fix: support object literal types in intersection member conflict resolution**
- **fix: recursively unwrap and walk inside type parameters of BuiltinTypes inside DependencyWalker**
- **fix: pass typeArg: true when resolving type parameters of referred BuiltinTypes and PackageWebTypes**
- **fix: propagate typeArg and parameter flags recursively inside UnionType and IntersectionType transformations**
- **fix: dynamically propagate and resolve specific type parameter bounds constraints inside ReferredType and getGenericTypes**
- **feat: add lines_longer_than_80_chars ignore rule to all generated interop files**
- **fix: escape JS properties/methods starting with _ to public Dart names starting with $**


js_interop_gen/lib/src/ast/base.dart
```diff
@@ -9,53 +9,77 @@ import 'documentation.dart';
...
+    this.nullable = false,
+    this.url,
+    this.isTypeArgument = false,
+    super.variadicArgsCount,
+    super.shouldEmitJsTypes,
+    super.redeclareOverrides,
+  });
 }
 
 class ASTOptions {
```
nikeokoronkwo: I think `ASTOptions` is stale. ~~I know we probably used to use it somewhere~~ This was meant to be somewhat a placeholder for parameters we would move to the config file, but I see it isn't being used anywhere, which is interesting. The generator runs successfully without this, so we should consider discarding this type.

js_interop_gen/lib/src/interop_gen/transform/dependency_walker.dart
```diff
@@ -0,0 +1,175 @@
...
+import '../../ast/declarations.dart';
+import '../../ast/types.dart';
+import '../transform.dart';
+
+/// A helper class that resolves the dependent declarations of a given AST
+/// [Node].
+class DependencyWalker {
+  /// Recursively gets all declarations that are referred to or needed by
+  /// the given [decl] node.
+  static NodeMap getDependenciesOfDecl(Node? decl, [NodeMap? context]) {
```
nikeokoronkwo: nit: maybe we could name this `extractDependencies` or `getDependencies`?

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -679,17 +720,39 @@ sealed class _UnionOrIntersectionDeclaration extends NamedDeclaration
 
   _UnionOrIntersectionDeclaration({
     required this.name,
-    this.types = const [],
+    List<Type> types = const [],
     List<GenericType>? typeParams,
   }) : typeParameters = typeParams ?? [] {
+    final uniqueTypes = <Type>[];
```
nikeokoronkwo, line 767: The reason I didn't dedupe types initially was because the typescript compiler should catch errors where multiple types of the same name are used anyways.

js_interop_gen/test/integration/interop_gen/generic_union_input.d.ts
nikeokoronkwo, line 1: Now I tried this:
```ts
export interface ComplicatedClass<T, U, V> {
    a: GenericContainer<T> | OtherContainer<U>;
    b: TargetClassWithGenericUnion<U | V>;
    c: T | U | V;
    d: GenericContainer<T | U > | OtherContainer<U | V>;
}
```

And the output failed to map the generic types from the interface to the parameter types:

```dart
extension type ComplicatedClass<
  T extends _i1.JSAny?,
  U extends _i1.JSAny?,
  V extends _i1.JSAny?
>._(_i1.JSObject _) implements _i1.JSObject {
  external AnonymousUnion_2676066 a;

  external TargetClassWithGenericUnion<AnonymousUnion_4191871> b;

  external AnonymousUnion_8475670 c;

  external AnonymousUnion_3234640 d;
}
```

The type parameters may be filtered out somewhere in the pipeline

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -56,7 +56,14 @@ class ReferredType<T extends Declaration> extends NamedType {
             ((declaration is NestableDeclaration)
                 ? (declaration as NestableDeclaration).completedDartName
                 : declaration.dartName ?? declaration.name)
-        ..types.addAll(typeParams.map((t) => t.emit(options)))
+        ..types.addAll(
+          typeParams.map((t) {
+            if (t == BuiltinType.$voidType) {
+              return BuiltinType.anyType.emit(options);
+            }
+            return t.emit(options);
```
nikeokoronkwo: nit: `return (t == BuiltinType.$voidType ? BuiltinType.anyType : t).emit(options)` here and in [R228-R231 of `js_interop_gen/lib/src/ast/builtin.dart` in the same commit](https://github.com/dart-lang/web/pull/549/changes/c0881a245464824cf82b9ea41dcd2e4ceb3393b7#diff-f776136a31f551bd152300d05bccc96183dd06b9246b72f8b3ba80aeee63221dR228-R231)

js_interop_gen/test/integration/interop_gen/string_literal_module_expected.dart
nikeokoronkwo, line 1: I do not think this should be the expected output of this... Based on how modules work, this could either be a scoped namespace with the same name, or a file prepended with a top-level library `@JS('vscode')` annotation. 

Modules work a bit weirdly from others, as they are largely dependent on how the JS code is coupled together with the Dart code and less about the interface, which is why they've been more difficult than other features in the generator because of how other declarations would use the modules if that makes any sense. See https://github.com/dart-lang/web/pull/446 for more.

js_interop_gen/test/integration/interop_gen/namespaces_expected.dart
```diff
@@ -55,7 +55,7 @@ extension type EnterpriseApp._(_i1.JSObject _) implements _i1.JSObject {
 external Data_Models_User get user1;
 typedef UserService = EnterpriseApp_DataServices_UserService;
 @_i1.JS()
-external UserService get userService;
+external EnterpriseApp_DataServices_UserService get userService;
```
nikeokoronkwo, line 59: The issue with this is that it now uses the flattened type instead of the `typedef` which it was before.

I'd assume we should go for one flow at least: either we make use of the typedefs, or we shake off typedefs not used

js_interop_gen/lib/src/ast/merger.dart
nikeokoronkwo, line 1: There are some issues with the deduplication system here:

- It doesn't dedupe between similar methods/constructors with optional parameters:
  ```ts
  interface Example {
    foo(x: string): void;
    foo(x?: string): void;
  }
  ```
- It doesn't dedupe between similar methods/constructors where one is a normal param and the other is rest
  ```ts
  interface Example {
    foo(...x: string[]): void;
    foo(x: string[]): void;
  }
  ```
- `readonly` properties
- `get`/`set` vs normal properties
- Generic type constraints
  ```ts
  interface Example {
    foo<T extends string>(x: T): T;
    foo<T extends number>(x: T): T;
  }
  ```
- static instances vs normal instances
- function parameter variance ignored

This may be a lot, so idk if we will have a better way to do this. We can update the IDs instead of having these methods to make it easier, but afaik the typescript compiler should catch duplicate declarations, unless a bug is present in our code

js_interop_gen/lib/src/ast/declarations.dart
```diff
@@ -1249,13 +1249,14 @@ class ConstructorDeclaration implements MemberDeclaration, Node<Constructor> {
   }
 
   @override
-  Constructor emit([covariant DeclarationOptions? options]) {
-    options ??= DeclarationOptions();
+  Constructor emit([covariant Options? options]) {
+    final declarationOptions =
+        (options ?? DeclarationOptions()) as DeclarationOptions;
```
nikeokoronkwo: Is there a reason we did this instead of just making the `options` parameter `DeclarationOptions` as it was before? It seems like the same thing is being done.

js_interop_gen/lib/src/interop_gen/transform/type_resolver.dart
nikeokoronkwo, line 1: I think we should move all or most of the type transformation into its own logic, as for the most part, the declaration transformer functions don't need to know the implicit stuff done in making a type and really only need the type itself. 

It would also help in terms of reducing the size of the `switch...case` statement we have into its own separate functions, and we could do more to refactor the current logic for type resolution (as you've already started)

js_interop_gen/test/integration/interop_gen/duplicate_references_expected.dart
```diff
@@ -0,0 +1,63 @@
+// ignore_for_file: camel_case_types, constant_identifier_names
+// ignore_for_file: library_private_types_in_public_api
+// ignore_for_file: non_constant_identifier_names, unnecessary_parenthesis
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'dart:js_interop' as _i1;
+
+extension type Thenable<T extends _i1.JSAny?>._(_i1.JSObject _)
+    implements _i1.JSObject {
+  external Thenable<AnonymousUnion_3555654>
```
nikeokoronkwo: This should return the type `Thenable<AnonymousUnion_3555654<TResult1, TResult2>>`. There should be a comment in a previous commit on this issue

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -773,7 +773,13 @@ sealed class _UnionOrIntersectionDeclaration extends NamedDeclaration
             return Method((m) {
               final word = _typeNameForGetter(t);
               final Expression body;
-              if (desugarTypeAliases(t) == repType) {
+              final jsAlt = jsTypeAlt;
+              final desugared = desugarTypeAliases(t);
+              if (desugarTypeAliases(t) == repType ||
+                  (jsAlt is BuiltinType && jsAlt.name == 'JSAny') ||
```
nikeokoronkwo: There should not be a case where there is a `NamedType` that isn't a `BuiltinType` called `JSAny` or `void`. In JS these are keywords so they should always be mapped to `JSAny`

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -54,6 +54,21 @@ class ReferredType<T extends Declaration> extends NamedType {
...
+      declTypeParams.addAll(decl.typeParameters);
+    } else if (decl is TypeAliasDeclaration) {
+      declTypeParams.addAll(decl.typeParameters);
+    }
+
+    final paddedTypeParams = [
+      ...typeParams,
+      if (typeParams.length < declTypeParams.length)
+        for (var i = typeParams.length; i < declTypeParams.length; ++i)
+          declTypeParams[i].constraint ?? BuiltinType.anyType,
```
nikeokoronkwo: I think type padding should not only be based on constraint but also default value, if possible. This is because if the type wasn't assigned it is inferred to be the default value, and not always its constraint alone.

What do you think?

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -78,10 +78,18 @@ class ReferredType<T extends Declaration> extends NamedType {
                 : declaration.dartName ?? declaration.name)
         ..types.addAll(
           paddedTypeParams.map((t) {
+            final typeArgsOptions = options == null
```
nikeokoronkwo: nit: move to top of func and use `options ??= TypeOptions()`

js_interop_gen/lib/src/ast/documentation.dart
```diff
@@ -24,6 +24,12 @@ class Annotation {
 
   Expression emit() {
     if (arguments.isEmpty) {
+      if (kind == AnnotationKind.deprecated) {
```
nikeokoronkwo, line 27: Is there a reason why we do this? Is Dart deprecating the `@deprecated` annotation?

js_interop_gen/lib/src/ast/documentation.dart
```diff
@@ -24,6 +24,12 @@ class Annotation {
...
+        return refer(
+          kind.name,
+          kind.source,
+        ).call([literalString('Deprecated')]);
```
nikeokoronkwo: It's fine if you leave this, but I was thinking of a "tad" more descriptive message like "This declaration has been marked as deprecated in the original JS/TS code" (obviously not the best of messages but an example of course)

js_interop_gen/lib/src/ast/declarations.dart
```diff
@@ -131,8 +131,13 @@ sealed class TypeDeclaration extends NestableDeclaration
           ),
     );
 
-    final repType = useFirstExtendeeAsRepType || this is ClassDeclaration
-        ? getRepresentationType(this)
+    final resolvedRepType = getRepresentationType(this);
+    final repType =
```
nikeokoronkwo, line 140: The reason we didn't allow this is because technically classes in JS can only be objects, so even if a TS class is extending a `String` TS interface for instance, its usages/instances should be compiled in JS as an object nonetheless (which is part of why it really isn't common practice to have a class extend something like `String` or `Number`). We can enforce this for builtin types that are ontop of `JSObject`, but that is a handful iirc.

js_interop_gen/test/integration/interop_gen/representation_type_inheritance_expected.dart
```diff
@@ -0,0 +1,48 @@
...
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'dart:js_interop' as _i1;
+
+extension type SpecialArray._(_i1.JSArray<_i1.JSString> _)
+    implements _i1.JSArray<_i1.JSString>, Object {
+  external double specialLength;
+}
+
+/// Provides functionality common to all JavaScript objects.
+extension type Object._(_i1.JSObject _) implements _i1.JSObject {
```
nikeokoronkwo: This should be a bug: we should map this to JSObject because they are supposed to be the same thing.

I understand this is not the scope of this PR, but I wanted to note this out incase.

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -792,6 +792,20 @@ sealed class _UnionOrIntersectionDeclaration extends NamedDeclaration
             if (method != null) memberDecls[m] = method;
           }
         }
+      } else if (e case ObjectLiteralType(
```
nikeokoronkwo, line 854: I don't think there's ever a case that an interface is extended by an object literal type:

Tried this in TS Online:
```ts
// erroneous TS Code
interface M extends { canRun: string; } {
  200: string;
}
```

And the error was: `An interface can only extend an identifier/qualified-name with optional type arguments.`

js_interop_gen/lib/src/ast/base.dart
```diff
@@ -92,13 +94,33 @@ abstract class NamedDeclaration extends Declaration {
...
-    typeParams: typeArgs?.toList() ?? [],
-    url: url,
-    isNullable: isNullable,
-  );
+  ]) {
+    final typeParamsList = typeArgs?.toList() ?? [];
+    final decl = this;
+    List<GenericType> declParams;
+    try {
+      final dynamic rawParams = (decl as dynamic).typeParameters;
```
nikeokoronkwo: nit: use of `dynamic`

js_interop_gen/test/integration/vscode_test_notes.md
nikeokoronkwo, line 1: nit: .md file

nikeokoronkwo (commented):
This is awesome. Thank you!

A few general notes:

* I noticed a lot of conflict resolution handling (i.e. deduping) in your code: technically we should not need to do most of that, either because at the point in the codebase it should be handled, or the TypeScript compiler should not allow such code to be generated or successfully pass through the parser anyways. We are not expecting people to write these `.d.ts` files so we should expect correct inputs.

cc @srujzs if you have the time to.

js_interop_gen/lib/src/ast/types.dart
kevmoo: Deduplication is intentional here to protect Dart from generating duplicate member getter names when emitting union extension types.

js_interop_gen/test/integration/interop_gen/namespaces_expected.dart
kevmoo: We currently emit typedefs for completeness, but we plan to explore pruning unused typedefs or directly referencing them in a future iteration.

js_interop_gen/lib/src/ast/merger.dart
kevmoo: Overload merging currently relies on structural equality; we are tracking ID-based overload disambiguation for a future architecture enhancement.

js_interop_gen/lib/src/ast/types.dart
kevmoo: You are completely right that an `interface` cannot extend an object literal in TypeScript (`interface M extends { foo: string }` is a syntax error).

However, this method (`_UnionOrIntersectionDeclaration._emit`) is shared between unions and **intersection types** (`_IntersectionDeclaration`). In TypeScript, intersection types frequently intersect named types with object literals (e.g. `type M = A & { foo: string; }`). 

When emitting an intersection extension type in Dart, `extendees` holds the component types of that intersection. Therefore, checking `if (e case ObjectLiteralType)` is essential here to unpack the properties and methods from object literals inside intersection types (`&`), even though they can never appear in interface `extends` clauses.

kevmoo:
@nikeokoronkwo PTAL 😄

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -726,8 +814,99 @@ sealed class _UnionOrIntersectionDeclaration extends NamedDeclaration
...
+                ..external = true
+                ..static = spec.static
+                ..returns = spec.returns
+                ..requiredParameters.addAll(spec.requiredParameters)
+                ..optionalParameters.addAll(spec.optionalParameters)
+                ..annotations.addAll(spec.annotations),
+            ),
+          );
+        } else if (spec is Field) {
+          conflictingMethods.add(
```
nikeokoronkwo, line 904: We should not forget to add necessary annotations from the `spec`. See this example:

```ts
export interface A {
    _priv: string;
    "dashed-prop": string;
}

export interface B {
    _priv: number;
    "dashed-prop": number;
}

export type Intersect = A & B;
```

In the current generated code, we do not get the `@JS` annotations for `_priv`

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -726,8 +814,99 @@ sealed class _UnionOrIntersectionDeclaration extends NamedDeclaration
...
 
+    final memberDeclCount = <String, int>{};
+    final memberDecls = <String, MemberDeclaration>{};
+
+    for (final e in extendees) {
+      if (e case ReferredType(declaration: final d) when d is TypeDeclaration) {
+        final members = getMemberHierarchy(d, true);
+        for (final m in members) {
+          memberDeclCount[m] = (memberDeclCount[m] ?? 0) + 1;
+          if (memberDecls[m] == null) {
```
nikeokoronkwo: nit: `memberDecls.containsKey(m)`

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -726,8 +814,99 @@ sealed class _UnionOrIntersectionDeclaration extends NamedDeclaration
...
+
+    final conflictingMembers = memberDeclCount.entries
+        .where((e) => e.value > 1)
+        .map((e) => e.key)
+        .toSet();
+
+    final conflictingMethods = <Method>[];
+
+    for (final m in conflictingMembers) {
+      final decl = memberDecls[m];
```
nikeokoronkwo, line 880: Conflicting declarations should have their (return/property) types resolved where such types differ. For instance:

```ts
export interface A {
    _priv: string;
    "dashed-prop": string;
}

export interface B {
    _priv: number;
    "dashed-prop": number;
}

export type Intersect = A & B;
```

In currently generated code, the type of `_priv` is assumed and output as `string`, while in actual TS, it should be `never`

js_interop_gen/lib/src/ast/helpers.dart
```diff
@@ -107,28 +122,31 @@ Set<String> getMemberHierarchy(
 }
 
 Type getRepresentationType(TypeDeclaration td) {
```
nikeokoronkwo, line 124: We should map some of the TS interfaces to actual types (like `Number`, `String`, etc) rather than generating them as if they are `JSObject` which isn't accurate in actual JS contexts. Classes that extend these types (like `Number`, `String`, etc), while being interfaces, should be `<symbol> & { /* rest */ }`. 

The one issue with this is if we want to deal with the ES spec, we'd have to *not* do this (somehow). Therefore, maybe we can have this as an option maybe?

nikeokoronkwo (commented):
Some more things...

---

Another issue is I ran this:

```ts
export interface SharedItem {
    value: number;
}

export declare namespace API {
    interface Container {
        item: SharedItem;
    }
}

// Re-export the same symbol that was referenced inside the namespace.
export { SharedItem as ReshapedItem };
```

And the generator did not expose `SharedItem` but instead only `ReshapedItem` (due to the export statement) and the API namespace.

js_interop_gen/lib/src/ast/merger.dart
nikeokoronkwo: So save this for a future issue/pr?

kevmoo:
take another look!

js_interop_gen/test/integration/interop_gen/project/output/a.dart
nikeokoronkwo, line 1: This should only produce the typedef if both the reference and the renamed symbols are exported at the same time. In this case, only the renamed symbol was exported, so the previous version was correct. This current one will fail because declarations like `PolarCoordinate`, `CylindricalCoordinate` and `SphericalCoordinate` are not exported.

The logic should produce the code prior to this.

js_interop_gen/test/integration/interop_gen/intersection_member_conflict_expected.dart
```diff
@@ -9,18 +9,38 @@ import 'dart:js_interop' as _i1;
...
+  external _i1.JSAny get conflictProperty;
+  external set conflictProperty(_i1.JSAny value);
+  @_i1.JS('_priv')
+  external _i1.JSAny get $_priv;
+  @_i1.JS('_priv')
+  external set $_priv(_i1.JSAny value);
+  @_i1.JS('dashed-prop')
+  external _i1.JSAny get dashedProp;
+  @_i1.JS('dashed-prop')
+  external set dashedProp(_i1.JSAny value);
```
nikeokoronkwo: nit: `_i1.JSAny?`

nikeokoronkwo: i.e. instead of `_i1.JSAny` for conflicting vars/prop types

kevmoo: Fixed!

js_interop_gen/test/integration/interop_gen/project/output/a.dart
kevmoo: Better?

js_interop_gen/lib/src/ast/builtin.dart
nikeokoronkwo, line 36: nit: do you think it is a good idea to annotate discardable instances with [`@doNotStore`](https://pub.dev/documentation/meta/latest/meta/doNotStore-constant.html)?

js_interop_gen/lib/src/interop_gen/transform/dependency_walker.dart
```diff
@@ -0,0 +1,175 @@
...
+import '../../ast/declarations.dart';
+import '../../ast/types.dart';
+import '../transform.dart';
+
+/// A helper class that resolves the dependent declarations of a given AST
+/// [Node].
+class DependencyWalker {
+  /// Recursively gets all declarations that are referred to or needed by
+  /// the given [decl] node.
+  static NodeMap extractDependencies(Node? decl, [NodeMap? context]) {
```
nikeokoronkwo, line 16: We might want to consider optimizing this function in terms of the number of `NodeMap` instances we create per recursive call, as this can become crazy when we deal with larger and more complicated .d.ts files.

nikeokoronkwo (approved):

nikeokoronkwo:
Can gemini review one last time?

js_interop_gen/lib/src/ast/builtin.dart
kevmoo: We could look at it?

js_interop_gen/lib/src/interop_gen/transform/dependency_walker.dart
kevmoo: done! Great catch!

js_interop_gen/lib/src/ast/builtin.dart
kevmoo: Gotcha! done!

js_interop_gen/test/integration/interop_gen/functions_input.d.ts
```diff
@@ -15,3 +15,7 @@ export declare function identity<T = string>(value: T): T;
 export declare function someFunction<A>(arr: A[]): undefined;
 export declare function someFunction<A, B>(arr: A[]): B;
 export declare function logTuple<T extends any[]>(...args: T): T;
+
+export interface ErrorReporter {
```
nikeokoronkwo: nit: can we move this example to some other place. I assume this is for testing just function declarations?

nikeokoronkwo (approved):

js_interop_gen/lib/src/ast/types.dart
```diff
@@ -264,19 +339,20 @@ class GenericType extends NamedType {
...
     return other is GenericType &&
         other.name == name &&
-        other.constraint == constraint;
+        other.constraint == constraint &&
+        other.defaultType == defaultType;
   }
 
   @override
-  int get hashCode => Object.hash(name, constraint);
+  int get hashCode => Object.hash(name, constraint, defaultType);
```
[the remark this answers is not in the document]
kevmoo: Hi! Thank you for the catch. We have updated `GenericType.id` to utilize the parent``s `qualifiedName` (or `name`) instead of the namer-suffixed `parent.id`. This ensures that the generated ID remains stable across multiple separate declarations of a merged generic interface, resolving signature-based member deduplication failures during AST merging. This change is fully implemented in commit `2faeaabec`!

js_interop_gen/lib/src/interop_gen/transform/type_resolver.dart
```diff
@@ -0,0 +1,631 @@
...
+            .map((TSNode decl) {
+              return transformer.programMap.getDeclarationRef(
+                importUrl,
+                decl,
+                fullyQualifiedName.asName,
+              );
+            })
+            .reduce(
+              (List<Node>? prev, List<Node>? next) => [...?prev, ...?next],
+            );
```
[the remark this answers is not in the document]
kevmoo: Hi! Thank you for the catch. Calling `.reduce` on an empty list throws a `StateError` at runtime. Since `declarations` can indeed be empty (if `getDeclarations()` returns null or an empty list), using `.reduce` was unsafe. We have replaced the `.reduce` call with a robust, high-performance `.expand().toList()` flattener that safely processes references and defaults to an empty list without crashing. This change is fully implemented in commit `2faeaabec`!

js_interop_gen/lib/src/interop_gen/transform/type_resolver.dart
```diff
@@ -0,0 +1,631 @@
...
+        if (cached != null) {
+          transformedDecls = cached;
+        } else {
+          transformedDecls = transformer.transformAndReturn(
+            declaration,
+            namer: namer,
+            parent: parent,
+          );
+        }
+
```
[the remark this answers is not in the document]
kevmoo: Hi! Thanks for the feedback. You are absolutely correct that `transformer.transformedCache` was being manually checked before calling `transformer.transformAndReturn`, which already encapsulates this exact cache lookup internally. We have removed the redundant manual check block to simplify the flow and improve code maintainability. This change is fully implemented in commit `2faeaabec`!

js_interop_gen/lib/src/interop_gen/transform/type_resolver.dart
```diff
@@ -0,0 +1,631 @@
...
+        }
+      }
+
+      return outputType;
+    }
+
+    // begin
+    final declarations = symbol!.getDeclarations()?.toDart ?? [];
+    if (declarations.firstOrNull?.kind == TSSyntaxKind.EnumMember) {
+      final enumDecl = (declarations.first as TSEnumMember).parent;
```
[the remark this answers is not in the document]
kevmoo: Hi! Thank you for the feedback. The `symbol` parameter is indeed typed as nullable (`TSSymbol?`), but it was immediately asserted with a bang `symbol!` and passed to other methods expecting a non-null symbol, creating a potential runtime crash vector. We have introduced an early null safety guard check at the beginning of the `getTypeFromSymbol` method, throwing a clear diagnostic exception if it is indeed null and allowing static type promotion. This change is fully implemented in commit `2faeaabec`!

js_interop_gen/test/integration/interop_gen/functions_input.d.ts
kevmoo: Hi Nike! Thank you for the catch. The `ErrorReporter` interface and its never-returning method are indeed interface declarations rather than top-level function declarations, so they did not belong inside the functions test suite. We have relocated the `ErrorReporter` test case to `interfaces_input.d.ts` (and updated its corresponding golden `interfaces_expected.dart`), keeping the functions test suite strictly focused on top-level function declarations. This is fully implemented in commit `a2901c9de`!

kevmoo:
Thank you @nikeokoronkwo !

PR #556, opened 2026-06-04
Remove DocumentCustomEvents

biggs0125:
Remove DocumentCustomEvents which now only contains duplicates of members in DocumentEventGetters.


kevmoo (approved):
CC @srujzs

srujzs (approved):
Thanks! I just got pinged on this so I'm reverting the DEPS change in the SDK to avoid the breakage.

I'm going to alphabetize things/standardize names a bit here so cases like this become more obvious.

PR #561, opened 2026-06-16
Only pass langVersion if snapshot

srujzs:
`dart bin/update_idl_bindings.dart` is currently broken when running locally because we use the web package path to determine the language version to run the generator. Because that's 3.10, use of private named parameters throws. We could pass the generator version here instead, but that logic is already in compileDartMain, so only pass the lang version if we're running from snapshot.

CI avoided this because it's presumably running from snapshot.


kevmoo (approved):

srujzs:
Thanks!

PR #562, opened 2026-06-22
[chore] Improve notice about updating the js_type_supertypes.dart

kevmoo:


js_interop_gen/lib/src/cli.dart
```diff
@@ -228,15 +229,19 @@ Future<void> checkJsTypeSupertypes() async {
             .replaceAll('\r\n', '\n')
             .replaceAll(sdkLineRegex, '')
             .trim()) {
+      final pinnedSdkMatch = RegExp(
+        r'^// Generated from Dart SDK (.*?)$',
+        multiLine: true,
+      ).firstMatch(currentContent);
```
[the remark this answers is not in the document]
kevmoo: Updated the RegExp pattern to use `(\S+)` and added a unit test to verify version extraction across LF and CRLF line endings. Thanks!

js_interop_gen/test/sdk_version_test.dart
```diff
@@ -62,4 +62,14 @@ void main() {
     check('3.12.0', Version(3, 12, 0));
     check('<4.0.0', Version(0, 0, 0));
   });
+
+  test('pinned SDK RegExp extracts clean version without carriage return', () {
```
srujzs: It feels like this is just testing `RegExp` itself and not any of our code. Delete?

srujzs (approved):

js_interop_gen/test/sdk_version_test.dart
kevmoo: Refactored to extract pinned SDK version parsing into an `extractPinnedSdkVersion` helper function in `sdk_version.dart`, and updated `sdk_version_test.dart` to test that package helper function directly.

